PR #1, opened 2016-08-04
Whitespace only changes to remove JS Standard warnings

melvincarvalho:
Fixes all JS Standard warnings for file :

    lib/acl-control.js

Confirmed as whitespace only:

    git diff -w


timbl:
Thanks @melvincarvalho !

PR #5, opened 2016-11-14
Initial People Picker

dan-f:
Hey @timbl - just making my efforts public here. Right now the picker can render a list of people dragged in via WebId URIs. I'm currently working on using the `kb` to keep track of the dynamic vcard group as we discussed last week.


dan-f:
Heads up that this change includes two refactors (which are currently causing conflicts which I can resolve):
1) Introduce babel
2) Begin to split apart the widgets module

PR #6, opened 2016-12-05
Stick icons in a static dir away from the build

dan-f:
@timbl, I hope this resolves the icon issues some have been seeing.

Icons now live in a `static/icons` directory, which is distributed as part of the `npm` package, which lets package consumers self-host those icons if they wish to.

In the long term, I don't think we should be using GitHub as a CDN for icons, but this should resolve issues once it merges and once the `gh-pages` branch is fast-forwarded to `master`.

There should be no reason to track the `lib` directory in git.


melvincarvalho:
Thanks for adding this, I've been having issues getting the icons to display lately.

melvincarvalho:
Gentle bump on this.  Did we decide whether to merge?

dan-f:
Hey @melvincarvalho thanks for the reminder. I floated it by @timbl but we didn't reach a decision. He had an idea to serialize the SVG icons within a JS module rather than serve each icon separately. There are also icon fonts such as [font awesome](http://fontawesome.io/) which I like because you only have to load a single resource for all your icons.

I think we should merge this to unbreak icons for the time being and then explore one of the other options as a follow-up item.

timbl:
1) When changing where icons are mounted it is important to leave them in the old place for a while as they will still be referenced from old code.
2) If we have to move them then maybe we should dedicate a URI which we can support indefinitely, like   solid.mit.edu/dist/  or   w3.org/somewhere...

dan-f:
I have a few ideas here. A starting point is to decouple the CDN from the resource paths. A basic implementation looks like:

```js
const cdn = base => path =>
  [base.replace(/(\/)$/, ''), path.replace(/^(\/)/, '')].join('/')

const githubIcons = cdn('https://linkeddata.github.io/solid-ui/static/icons/')
const iconPath = githubIcons('noun_144.svg') // => https://linkeddata.github.io/solid-ui/static/icons/noun_144.svg
// or
const myCdn = cdn('https://mycdn.biz')
const icon = myCdn('static/icons/noun_144.svg')
const picture = myCdn('static/img/this_does_not_exist_but_could.jpg')
```

At this point the library can support a sort of cdn dependency injection. Every call to a function which requests static assets from a CDN has an optional argument, defaulting to, e.g., `cdn('https://cdn.w3.org/solid')`. The problem with this approach is that anyone using their own CDN would have to pass the cdn function around everywhere.

A better approach, I think, is to have a global variable for the CDN, and expose it during the build process using a tool like [webpack](https://webpack.js.org/). Here's [an example](https://github.com/dan-f/mark/blob/master/webpack.prod.config.js#L7-L9) of this type of strategy.

Then our code can look like this:
```js
// in cdn.js
export default function cdn (path) {
  // SOLID_UI_CDN is templated at build time
  const cdnBaseUrl = SOLID_UI_CDN || 'https://cdn.w3.org/'
  return [cdnBaseUrl.replace(/(\/)$/, ''), path.replace(/^(\/)/, '')].join('/')
}
```
```js
// in consumer code
import cdn from './cdn'

function renderThing () {
  // ...
  const iconUrl = cdn('icons/noun_144.svg')
  // ...
}
```

dan-f:
And I agree that we should leave the icons where they are for a while during the transition.

timbl:
(I agree that we are not tracking icons in li directory. They are are only in the src  directory, where they were moved to when src was set up, and babel -d src lib .  So the iconbase had already been moved from lib to src.)     

Yes agree about something like that for the cdn.
Note there is something like cdn global variable, in $SolidTestEnvironment in src/iconbase.js:
```
  if (typeof $SolidTestEnvironment !== 'undefined' && $SolidTestEnvironment.iconBase) {
    module.exports.iconBase = $SolidTestEnvironment.iconBase;
    module.exports.originalIconBase = $SolidTestEnvironment.originalIconBase;
  } else {
    module.exports.iconBase = iconsOnGithub + '/icons/';
    module.exports.originalIconBase = iconsOnGithub + '/originalIcons/';
  }
```
There is a case for allowing the CDN to be overridden either before the library is called, or after, just because will vary in how they need to do this.  Here for example you can et up $SolidTestEnvironment in advance, or you can overwrite the value of UI.iconBase later...

PR #12, opened 2017-08-15
Timbl   Make URI input tolerate no workspaces; extract newThingUI to this layer

timbl:
This provides the solid-ui level support for having consistent new thing creation objects in different panes, including the folder pane (see timbl-folders branch of solid-app-set)


dmitrizagidulin:
Conflicts resolved, looks ok, merging.

PR #17, opened 2017-09-14
Update solid-auth-client to 1.0.0

RubenVerborgh:


src/signin.js
```diff
@@ -702,9 +702,8 @@ function signInOrSignUpBox (myDocument, setUserCallback) {
     let idpUri = document.getElementById('idpInput')
 
     return solidAuthClient.popupLogin({ popupUri: config.popupUri})
-      .then(session => {
-        let webIdURI = session.webId
-        setUserCallback(webIdURI)})
+      .then(webIdFromSession)
+      .then(webIdURI => setUserCallback(webIdURI))
```
RubenVerborgh: I suppose this can just be `setUserCallback`.

dan-f: yep.

PR #19, opened 2017-11-21
Timbl acl control

timbl:
Fixing the new auth methods


src/signin.js
```diff
@@ -93,39 +99,67 @@ function saveUser (webId, context) {
...
 
+/*  Stop storing it in prefs now we have localStorage
   let webId = tabulator.preferences.get('me')
-
   return webId ? $rdf.sym(webId) : null
+  */
+  return null
+}
+
+/** Checks syncronously whether user is logged in
```
RubenVerborgh: Can this always happen synchronously? Might be better to have this async to be future-proof.

src/signin.js
```diff
@@ -789,15 +799,15 @@ function checkUser (setUserCallback) {
...
- * @param myDocument
+ * @param dom
  * @param listener
  *
  * @returns {Element}
  */
-function loginStatusBox (myDocument, listener) {
-  var me = currentUser()
+function loginStatusBox (dom, listener) {
+  var me = defaultTestUser()
```
RubenVerborgh: How come there is a test user here?

src/signin.js
```diff
@@ -888,7 +898,7 @@ function selectWorkspace (dom, appDetails, callbackWS) {
   var noun = appDetails.noun
   var appPathSegment = appDetails.appPathSegment
 
-  var me = currentUser()
+  var me = defaultTestUser()
```
RubenVerborgh: Idem.

PR #20, opened 2018-01-14
Standard format, tweak to chat, and readme

timbl:
This is a hard-to-review set of standard --fix changes. 
A small change to the chat pane.
Added some README list of features.


timbl:
Merging as it is almost just feross/standard syntax change

PR #26, opened 2018-03-01
Timbl prefernces

timbl:
Error handling around prefs
Bug fix in outlinemode link


timbl:
Should improve handling of preferences/type index file and specifically stop it trying to create a new blank file when the old one exists but there was an error getting it.

PR #29, opened 2018-03-14
rename and export logInLoadPreferences, export logInLoadProfile

timbl:


timbl:
Merging as NPM refernce to git branches seems broken and so pushing to master and to npmjs

PR #31, opened 2018-04-28
Add pushstate to forward links closes #30

### LINKED ISSUE #30: Address bar does not change when you click a link
When clicking a link in solid-ui the address bar retains the old URI.

Most browser do not allow a change or origin from javascript for security reasons.

However, for links in the same origin the history API pushstate function can be used

Example code:

```JavaScript
  if (UI.outline){
    UI.outline.GotoSubject(UI.store.sym(uri), true, undefined, true, undefined)
    if ( history.pushState ) history.pushState( {}, uri, uri );
  } else if (window && window.UI && window.UI.outline ){
    window.UI.outline.GotoSubject(UI.store.sym(uri), true, undefined, true, undefined)
    if ( history.pushState ) history.pushState( {}, uri, uri );
  } else {
    console.log("ERROR: Can't access outline manager in this config")
  }
```

at : https://github.com/solid/solid-ui/blob/master/src/widgets/index.js#L1104

melvincarvalho:
Partial solution that fixes clicking through

Does not address cross origin clicks or the back button, and possibly some other use cases

### DIFF
--- a/src/widgets/index.js
+++ b/src/widgets/index.js
@@ -1115,8 +1115,10 @@ UI.widgets.openHrefInOutlineMode = function (e) {
   const dom = window.document
   if (dom.outlineManager) {
     dom.outlineManager.GotoSubject(UI.store.sym(uri), true, undefined, true, undefined)
+    if (window.history.pushState) window.history.pushState({}, uri, uri)
   } else if (window && window.panes && window.panes.getOutliner) {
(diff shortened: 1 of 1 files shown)


melvincarvalho:
Done

melvincarvalho:
After [discussion](https://gitter.im/solid/chat?at=5ae70c64dad6fb186f002585), better to put this in [GotoSubject](https://github.com/solid/solid-panes/blob/master/outline/manager.js#L1628)

PR #36, opened 2018-05-26
Use preference values as options in widget

timbl:
Better than without.  Issues exist with form checkbox patches not being able to turn a true into a false, but may be a server issue.

### DIFF
--- a/src/infiniteMessageArea.js
+++ b/src/infiniteMessageArea.js
@@ -27,7 +27,8 @@ module.exports = function (dom, kb, subject, options) {
 
   options = options || {}
 
-  var newestFirst = !!options.newestFirst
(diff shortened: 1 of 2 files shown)


src/infiniteMessageArea.js
```diff
@@ -274,9 +275,13 @@ module.exports = function (dom, kb, subject, options) {
     renderMessage(messageTable, bindings, messageTable.fresh) // fresh from elsewhere
   }
 
-  function elementForImageURI (imageUri) {
+  function elementForImageURI (imageUri, options) {
     let img = dom.createElement('img')
-    img.setAttribute('style', 'max-height: 10em; border-radius: 1em; margin: 0.7em;')
+    let height = '10'
```
melvincarvalho: Minor nit, magic number.  I'm OK with this tho.

PR #41, opened 2018-08-24
Signin: Fix login/logout control to restore logout button.  Chat: inline URIs linked

timbl:


RubenVerborgh:
You might be interested in waiting for solid-auth-client 3.0.0, which is planned for today. It has a `trackSession` method, which allows for easily synchronizing state across all buttons on a page.

Here's an example login/logout button implemented in React:
https://github.com/solid/solid-auth-client/blob/bd30187e60b133f3d57fbe11c2a0593c019e01c9/demo/components/LoginButton.js

src/signin.js
```diff
@@ -888,30 +894,48 @@ function loginStatusBox (dom, listener) {
...
+      }
+      box.me = me ? me.uri : null
+    }, err => {
+      alert('loginStatusBox: ' + err)
+    })
+  }
+
+  if (solidAuthClient.trackSession) {
+    solidAuthClient.trackSession(session => {
+      if (session && session.webId) {
```
RubenVerborgh: If `session` exist, then it will have `webId`, so just `if (session)` is okay.

src/signin.js
```diff
@@ -888,30 +894,48 @@ function loginStatusBox (dom, listener) {
...
+        me = $rdf.sym(session.webId)
+      } else {
+        me = null
+      }
```
RubenVerborgh: Or if you want it fancy:
```
me = session ? $rdf.sym(session.webId) : null
```

src/signin.js
```diff
@@ -888,30 +894,48 @@ function loginStatusBox (dom, listener) {
...
+      box.refresh()
+    })
   }
```
RubenVerborgh: My only worry here is whether you will not end up with a double-refresh, i.e., should the other refresh code be put in the `else` statement? (i.e., trackSession does not exist)

RubenVerborgh:
solid-auth-client v2.2.0 published (minor version bump, since the API remained fully compatible)

timbl:
I’m thinking about merging https://github.com/solid/solid-ui/pull/41 as the fact that is tolerant of not being able to load preference file errors because of cross-site access problems is quite important.  Things which ask for type indexes used to crash out  or give alerts … when in fact this is common and we just hae to suck it up (and give up having user-specific preferebces and use private lists of good things).

PR #42, opened 2018-08-30
Update solid-auth-client to v2.2.3

RubenVerborgh:


timbl:
Overtaken by events. solid-auth-client is already  pulled in at 2.2.6

PR #49, opened 2018-10-21
Timbl bugfix - fix drag and drop bug in ACL control

timbl:


timbl:
Thanks for the review.. yes the ACL control acl-control.js  is a big thing, but the non IU parts have been split out in acl.js.  The complexity of it largely comes from the fact hat while the logic is about Authorization objects, the user we decide in fact thinks better in terms of combinations of mode "combos". These get different colors, and have made up names etc, and in fact even turn out to be  a nice way of serializing the ACL.     The non-separation of logic and style aand logic is in a way a bit of a a counterculture, yes.   But it makes the system modular.  Because the control supplies al its own styles, it will always work wherever you slap it into some other system, and it doesn't need any coordination with global style sheet conventions.  The aim is to get rid of the global style sheet. After all.. it is a global, with ll the problems which globals have: clashes of names, coordination of naming conventions,  unintended consequences, and so on.

PR #54, opened 2018-11-07
Remove black background from anonymous pad viewing. Fixes #52

### LINKED ISSUE #52: Notepad is displaying black font with a black background = black box
I have created a Notepad in my Public folder. As an authenticated user I can input and save as normal, and it all looks fine. But viewing from an unauthenticated user my notepad content looks like a big black box. The reason is that the font is black (colour #222)and the background is also black (colour #111). If I change the background colour to #ffffff white, I can then see my text.

Example: https://dontai.inrupt.net/public/Introduction/
There is text on lines 1,2,4,and 6. Highlight any of these lines and you will see the text.

solid-notepad-black-lines-2018-10-01 23-00-45-cropped

After around 10 minutes the background colour changes to background-color: #fed0f7; and text colour color: black; , and I can see the text. These new colours match what I see if I am authenticated and can edit the document. After 10 minutes the text and background colours are completely different from when I first display the page. The page changes colours by itself and without me touching the window. This is very odd.

solid-notepad-black-lines-after-10-mins-2018-10-05 08-01-04

JordanShurmer:


TorontoBoy:
Issue remains the same. I am unauthenticated, 
try https://dontai.inrupt.net/, public, document is "Introduction".

On both I can faintly see my text in a dark grey. I can highlight a line and see the text.
On Firefox I see a black box, then after 10 minutes I see my text black font with a pink background. I do not know why it automatically changes colour. On Opera I see a black box and it stays that way, no change after 10 minutes.

megoth:
Yeah, I'm about to check this on my machine as well, but I suspect the background color is bad, as it's very low contrast between text and background. Maybe set background-color to #fff instead?

megoth:
Another note on this: the visitor cannot edit the notes, so this shouldn't be an input at all. I would change the logic of this to actually create a paragraph or div. This could of course have a slew of other side-effects I haven't thought of though, so approach with care.

JordanShurmer:
The code change in this PR is not deployed to that website @TorontoBoy, or at least it's not right now. If I map the JS to my local build of this PR then it works as expected:

image

@megoth  - this solution in this PR is to remove the background-color entirely, which lets it use the browser default. I suppose we could set it to #fff instead, but that's not typical.

megoth:
> @megoth - this solution in this PR is to remove the background-color entirely, which lets it use the browser default. I suppose we could set it to #fff instead, but that's not typical.

Yeah, there's probably better solutions, but I would argue those involve not manipulating CSS with JS in this way ;)

Thank you for the commit though ^_^

TorontoBoy:
When will this new change be implemented on inrupt? The issue still exists for me, so I cannot confirm the fix.

kjetilk:
Hmmm, I think it might take a few days, @TorontoBoy . We should roll a release soon, and then get that release up on the server. Actually, I think I should be able to roll the release tomorrow morning!

PR #56, opened 2018-12-02
Timbl bugfix2 - signin

timbl:
clean up ns, add one, fix context.indexes typo in signin

### DIFF
--- a/src/ns.js
+++ b/src/ns.js
@@ -1,54 +1,9 @@
-// Namespaces we commonly use and have acommon prefixes for around solid
+// Namespaces we commonly use and have common prefixes for around Solid
 //
-var thisModule = {}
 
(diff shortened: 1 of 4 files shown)


src/ns.js
```diff
@@ -1,54 +1,9 @@
-// Namespaces we commonly use and have acommon prefixes for around solid
+// Namespaces we commonly use and have common prefixes for around Solid
 //
-var thisModule = {}
 
-module.exports = thisModule
+const solidNamespace = require('solid-namespace') // Delegate to this which takes RDFlib as param.
```
melvincarvalho: is solid-namespace in package.json?

timbl: Good catch . fixed

PR #64, opened 2019-05-22
Pass the store when creating a new pane

Vinnl:
When panes are loaded dynamically and thus are in their own Node module with their own scope, they do not have access to the global `UI` object, and therefore also cannot access the store. Thus, we should pass the store to the `mintNew` function when it is provided by a dynamically-loaded pane.

(@megoth It appears I also don't have rights to this repository yet.)


Vinnl:
Given that dynamic panes has been put on hold, I'm closing this for now.

PR #65, opened 2019-05-31
Fixing bug in regex for phone field

megoth:
The regex for phones is still not great, but we'll leave it at this for now until we have a more thorough change of the profile pane

### DIFF
--- a/src/widgets/index.js
+++ b/src/widgets/index.js
@@ -824,7 +824,7 @@ UI.widgets.fieldParams[UI.ns.ui('TextField').uri] = { }
 
 UI.widgets.fieldParams[UI.ns.ui('PhoneField').uri] = { 'size': 20, 'uriPrefix': 'tel:' }
 UI.widgets.fieldParams[UI.ns.ui('PhoneField').uri].pattern =
-  /^\s*\+?[ 0-9-]+[0-9]\s*$/
+  /^\s*\+?[0-9-]+[0-9]\s*$/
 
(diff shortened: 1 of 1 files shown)


src/widgets/index.js
```diff
@@ -824,7 +824,7 @@ UI.widgets.fieldParams[UI.ns.ui('TextField').uri] = { }
 
 UI.widgets.fieldParams[UI.ns.ui('PhoneField').uri] = { 'size': 20, 'uriPrefix': 'tel:' }
 UI.widgets.fieldParams[UI.ns.ui('PhoneField').uri].pattern =
-  /^\s*\+?[ 0-9-]+[0-9]\s*$/
+  /^\s*\+?[0-9-]+[0-9]\s*$/
```
Vinnl: Shouldn't we also disallow whitespace on either side of the phone number?

Vinnl:
I'm fine with merging, so go ahead if you want - just not sure if whitespace at the start or end of a phone number won't cause problems either :)

src/widgets/index.js
RubenVerborgh: While we're at it, shall we just `\d`?

megoth: Can do - wanted to keep the changes minimal, as to not introduce any errors (I'm not very good with regex, after all). But I've tried a slight change now, hopefully to the better.

megoth:
> I'm fine with merging, so go ahead if you want - just not sure if whitespace at the start or end of a phone number won't cause problems either :)

Tried to improve the regex used - please double-check, then we'll start publishing updates to NPM.

PR #67, opened 2019-05-31
Added watch-script

megoth:
Another small script to make life a bit easier for developers


Vinnl:
I changed it to use the long form of the arguments so I can see what it's doing without looking up the documentation. Otherwise LGTM!

PR #70, opened 2019-06-04
69 introduce typescript

### LINKED ISSUE #69: Introduce TypeScript
Since solid-ui is a library that _is_ used by multiple packages, it makes sense to provide type definitions, to make it easier for consumers to figure out how to use it.

Vinnl:
Builds on #68. This keeps compiling files with Babel like before, but additionally runs `tsc` to generate `.d.ts` files for TypeScript files.

I believe this should work just like before, but... I can't load my local copy of solid-ui anymore, whether this branch or `master`, so I can't really test it :(

Fixes #69.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -63,6 +63,6 @@ const UI = {
 }
 
 if (typeof window !== 'undefined') {
-  window.UI = UI  // Simpler access by non-node scripts
+  (window as any).UI = UI  // Simpler access by non-node scripts
 }
 module.exports = UI
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,13 +31,15 @@ https://github.com/solid/solid
  * @main solidUi.js
  */
 
+
+import rdf from 'rdflib'
+import ns from './ns'
+import style from './style'
+
 /**
  * @class SolidUi
  * @static
  */
-
-const rdf = require('rdflib') // pull in first avoid cross-refs
-const ns = require('./ns')
 
 const UI = {
   ns,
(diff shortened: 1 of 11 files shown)


src/index.ts
```diff
@@ -48,21 +54,23 @@ const UI = {
   create: require('./create'),
   icons: require('./iconBase'),
   log: require('./log'),
-  matrix: require('./matrix'),
+  matrix: matrix,
```
megoth: Maybe just use the short hand form, to clean up the code up a bit more?

src/preferences.ts
```diff
@@ -78,42 +86,47 @@ function recordPersonalDefaults (klass, context) {
...
-    }, err => {
+    }, (err: Error) => {
       reject(err)
     })
   })
 }
 
-function renderPreferencesForm (subject, klass, preferencesForm, context) {
+function renderPreferencesForm (
+  subject: $rdf.NamedNode,
```
megoth: Isn't it better with simply `NamedNode`?

src/index.ts
Vinnl: I'm generally not a fan of mixing the shorthand and regular form (or of using the shorthand form outside of one-liners in general), but I don't feel strongly about it so if you do - sure :) In that case, might also be a good idea to add an ESLint rule?

src/preferences.ts
Vinnl: Haven't heard of that before, why? Should I change the other `$rdf.*` imports as well?

megoth:
Closing this in favor of https://github.com/solid/solid-ui/pull/127

PR #71, opened 2019-06-06
ACL control adds web app control; Chat

timbl:


megoth:
I rebased the branch chat, and fixed some conflicts.

This PR is too big to give a proper review, but the code changes that I see look reasonable. There might be some bugs in this, but I think they will become easier to spot and remove once we started the work on migrating to TypeScript (which is happening soon).

Issue #76, opened 2019-07-25
Need to support acl:default


megoth:
`acl:defaultForNew` is obsolete, but the code in this project does not reflect this. We should check for both predicates to be backward-compatible.

megoth:
Implemented in https://github.com/solid/solid-ui/commit/cbefd1398467d1ee81c1ef2914a7d3245a9482d4

PR #77, opened 2019-07-25
Changing from defaultForNew to default with fallback

megoth:
The fallback strategy is to query for `acl:defaultForNew` (in addition to `acl:default`), to handle resources that uses the obsolete property.

Making sure to use `acl:default` for new resources.

This fixes https://github.com/solid/solid-ui/issues/76


megoth:
This was done in another commit (https://github.com/solid/solid-ui/commit/cbefd1398467d1ee81c1ef2914a7d3245a9482d4), so will close this PR.

Issue #78, opened 2019-08-05
`tryParent` in acl.js does not stop before trying to fetch https://


megoth:
`tryParent` is a recursive function that takes a URL (e.g. https://foo.com/bar/baz) and tries to find an ACL for that resource by looking at its parent if none are found initially (e.g. https://foo.com/bar/ --> https://foo.com/). 

There are a couple of problems with this:

1) Right now there is a bug that makes it also check https://. We need to stop the recursive function before this happens and give the appropriate error. (There should always be a root ACL, but errors might occur still.)
2) It will call tryParent if no triples for `acl:default` are found (https://github.com/solid/solid-ui/blob/master/src/acl.js#L359). This would mean that even though an ACL resource are found, the code will still look for another parent ACL resource, which I think goes against point 2 in https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm.

PR #79, opened 2019-08-05
Stop try parent on HTTP STATUS 200

megoth:
This should solve https://github.com/solid/solid-ui/issues/78

### DIFF
--- a/src/acl.js
+++ b/src/acl.js
@@ -327,7 +327,6 @@ UI.acl.setACL = function (docURI, aclText, callbackFunction) {
 UI.acl.getACLorDefault = function (doc, callbackFunction) {
   UI.acl.getACL(doc, function (ok, status, aclDoc, message) {
     var kb = UI.store
-    var ACL = UI.ns.acl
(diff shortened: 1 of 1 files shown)


timbl:
Hmmm

timbl:
tryParent needs to check and exit if there is no acl file atthe root of the domain, to avoid chopping the domain name off

timbl:
https://github.com/solid/web-access-control-spec/blob/master/README.md
says 

> Failing that, move up the container hierarchy until you find a container with an existing ACL file, which has some permissions to inherit.

TallTed:
Have to remember that "move up the container hierarchy" doesn't necessarily mean "truncate the path, solidus by solidus" as the URI's PATH element may not (though it often will) match the logical container hierarchy.  Challenge deepens as Containers identify the Members they contain, but Members don't necessarily identify their Containers.

megoth:
This was handled in https://github.com/solid/solid-ui/pull/80, so closing this PR.

PR #80, opened 2019-08-05
Alternative solution to https://github.com/solid/solid-ui/issues/78

megoth:


megoth:
> Being finicky, as we have that test ow twice on each recursive call to tryParent, maybe it should e replaced by a single test at the beginning of tryParent, before the getACL .

Right, I thought it would work since it was the logic being done already at https://github.com/solid/solid-ui/blob/f09f540bbcedbeb8085d7c25b65908ae48909820/src/acl.js#L348 . But yes, We should do that test before making the call.

> In fact the test `left >= right` will catch `https://` (left and right will both be 8) but won't catch `https://foo.com/` where left by 8 and right 16. So the right test was in the wrong place?

But `https://foo.com/.acl` might also be a ACL resource we want to check? I.e. when a user has a single user Pod server setup that serves their domain? Maybe I misunderstand you, but I think the test `left >= right` is the only one we need to check for?

megoth:
@timbl I've changed the code based on your feedback now.

I also found a place where you had forgotten to add the `ACL('default')`-fix.

Lastly, I added a TODO for you to review - didn't really find much use of that method in the projects I have in my current setup (NSS, mashlib, solid-panes, solid-ui, rdflib), so uncertain of it's usage. Might be a method we want to mark as obsolete and remove at next major release?

megoth:
> Beware of "^0.21.1" for rdflib maybe ">=0.21.1" ... maybe personal preference

I think this is how NPM does it by default, so didn't think more of it.

PR #91, opened 2019-08-14
Only adding a slash to destUri if folderName does not already have it

megoth:
The handling of ACL for root failed because the URI to PUT the file to contained an extra slash (e.g. `https://test.com//foo.txt` instead of `https://test.com/foo.txt`). This was an error that also happens when trying to upload files elsewhere, but it was allowed then.

This should fix https://github.com/solid/solid-panes/issues/143


megoth:
Let's discuss the IE-thing on the upcoming standup. We need to document this somewhere as well.

PR #93, opened 2019-08-15
Do not prefill a value for a checkbox

Vinnl:
If the BooleanField in the given resource is not set, the tristate
logic would set it to false by default. This commit changes that to
differentiate between the value that is shown (false), and the
value that's actually present on the web.

These changes were made as per Tim's instructions. They came up when working on https://github.com/solid/solid-panes/issues/144.


megoth:
I'll fix a new version that does not conflict with master and then merge it ^_^

PR #95, opened 2019-08-16
Readonly forms

timbl:
Where the data being viewed by a form is not editable -- either server capability or access control -- this PR disables user input, making a read-only view which is is cleaner, and does not invite editing.

### DIFF
--- a/src/signin.js
+++ b/src/signin.js
@@ -998,6 +998,7 @@ function loginStatusBox (dom, listener, options) { // 20190630
   }
 
   var logoutButton = function (me, options) {
+    options = options || {}
     const signInButtonStyle = options.buttonStyle || getDefaultSignInButtonStyle()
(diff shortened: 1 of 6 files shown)


timbl:
rebased

src/signin.js
```diff
@@ -998,6 +998,7 @@ function loginStatusBox (dom, listener, options) { // 20190630
   }
 
   var logoutButton = function (me, options) {
+    options = options || {}
```
megoth: Not important, but you could've done this in the previous line as well, with `function (me, options = {})`

megoth:
I'll merge this to master after we've done a publish of the latest rdflib, as this is dependent on the latest changes there.

megoth:
(Will probably release new version of rdflib on Sunday)

Issue #94, opened 2019-08-15
Stop specific sharing pane on /(root) deletes .acl


bourgeoa:
on NSSv5 and sure for v5.1.6 and v5.1.7: 
- `Stop specific sharing pane on /public` deletes `.acl` and your cannot edit it anymore (404).
- If you try to go to `/(root)` and `Set the sharing of folder content`, it does not recreate .acl in /public

In fact you can even **Stop specific sharing pane on /(root) it deletes .acl and you are out of your pod**

>  No ACL found for https://alain.bourgeoa.ga:8517/, searched in 
>  - https://alain.bourgeoa.ga:8517/.acl

I already submitted this issue # 1257 on solid/node-solid-server on july, 6

megoth:
Aoch, good find! We should try to at least prevent that from being done with the data browser. But yeah, need to be programmatically impossible on the server as well, good call ^_^

bourgeoa:
- I looked into the code and it seems to me that a way to do it is `acl-control.js` not to display the `Stop specific sharing for this` by including after L584 a :

> test that 
-- if( ! ( targetACLDoc.uri.endsWith('/.acl') && ( is in /(root) || is in first level folder (/public, /settings, ...) ) ))

- The same can be done in node-solid-server in `Idp.js` at `function delete`, but it does not avoid the use of `fetch(targetACLDoc.uri, options(DELETE ...))` which is what rdflib.js uses.

megoth:
Fixed in https://github.com/solid/solid-ui/pull/96

PR #96, opened 2019-08-19
Prevent users from deleting their root ACLs

megoth:
This checks whether pathname indicates that the target is root, or whether the root ACL itself states that the target is root.

Going to create a PR for NSS as well that adds `<.> a sp:Storage.` as default for all _new_ root ACLs.

### DIFF
--- a/src/acl-control.js
+++ b/src/acl-control.js
@@ -612,14 +612,17 @@ UI.aclControl.ACLControlBox5 = function (subject, dom, noun, kb, callback) {
           var q = str.indexOf('//')
           var targetDocDir = ((q >= 0 && p < q + 2) || p < 0) ? null : str.slice(0, p + 1)
 
-          if (targetDocDir) {
+          // eslint-disable-next-line no-undef
+          const targetIsStorage = kb.holds(targetDoc, UI.ns.rdf('type'), UI.ns.space('Storage'), targetACLDoc) || (location && location.pathname === '/')
+
+          if (!targetIsStorage && targetDocDir) {
             UI.acl.getACLorDefault($rdf.sym(targetDocDir), function (ok2, p22, targetDoc2, targetACLDoc2, defaultHolder2, defaultACLDoc2) {
               if (ok2) {
                 prospectiveDefaultHolder = p22 ? targetDoc2 : defaultHolder2
               }
               addDefaultButton(prospectiveDefaultHolder)
             })
-          } else {
+          } else if (!targetIsStorage) {
             addDefaultButton()
           }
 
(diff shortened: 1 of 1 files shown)


megoth:
This fixes https://github.com/solid/solid-ui/issues/94

src/acl-control.js
```diff
@@ -612,14 +612,19 @@ UI.aclControl.ACLControlBox5 = function (subject, dom, noun, kb, callback) {
           var q = str.indexOf('//')
           var targetDocDir = ((q >= 0 && p < q + 2) || p < 0) ? null : str.slice(0, p + 1)
 
-          if (targetDocDir) {
+          // @@ TODO: The methods used for targetIsStorage are HACKs - it should not be relied upon, and work is
+          // @@ underway to standardize a behavior that does not rely upon this hack
+          // eslint-disable-next-line no-undef
```
Vinnl: Is the `no-undef` disabling due to `UI`? Might be better to tell eslint that it's an available global, then.

(And if it's about `location`, perhaps best to use `window.location`.)

src/acl-control.js
```diff
@@ -612,14 +612,19 @@ UI.aclControl.ACLControlBox5 = function (subject, dom, noun, kb, callback) {
           var q = str.indexOf('//')
           var targetDocDir = ((q >= 0 && p < q + 2) || p < 0) ? null : str.slice(0, p + 1)
 
-          if (targetDocDir) {
+          // @@ TODO: The methods used for targetIsStorage are HACKs - it should not be relied upon, and work is
+          // @@ underway to standardize a behavior that does not rely upon this hack
```
Vinnl: Ideally, you'd also include a link here to where one can check the status of the work that is underway, so that engineers coming across this in the future can tell whether the hack can be removed. (And for bonus points: a note about what the risks of the current hack are.)

bourgeoa:
it solves `/(root)` but not avoid /public to delete is `/public/.acl` with the consequence of being unable to rebuild/edit `/public/.acl` in the future (see also issue #99)

megoth:
@bourgeoa thanks for finding this. So this seems to be a bit more complex issue than we initially thought - I will do a bit more research to see how this can be mitigated. I'll update this thread once I know more.

megoth:
@bourgeoa We don't need to protect the folders under root, but https://github.com/solid/solid-ui/issues/99 shows us that we have a bug, namely that we currently cannot create new .acl files for folders directly under root.

I will for now only protect root ACL, and work on https://github.com/solid/solid-ui/issues/99 to find a solution that allows for creating new .acl files for folders directly under root.

megoth:
@timbl I've added the changes that you proposed.

Please review again, and merge if everything looks good.

PR #109, opened 2019-08-28
Don't store a field's default value in the store

### LINKED ISSUE #105: 409 Error when attempting to edit Profile Highlight Color
If you attempt to edit the Highlight Color on your Profile, you get an error:
`Web error: 409 (Conflict) on PATCH of <https://khoward5.dev.inrupt.net/profile/card>`

## To Replicate
1. Log into a dev.inrupt.net Pod.
2. "Edit your profile".
3. Edit the Highlight Color. 

## Expected Behaviour
1. The Highlight color is updated to the selected color.

## Actual Behaviour
1. Error `Web error: 409 (Conflict) on PATCH of <https://khoward5.dev.inrupt.net/profile/card>` is displayed.

Vinnl:
Tim has said in a call that it's weird that this line is there in the first place and should be removed, so I guess this doesn't need another review from him.

Fixes #105.

### DIFF
--- a/src/widgets/forms.js
+++ b/src/widgets/forms.js
@@ -362,7 +362,6 @@ forms.field[UI.ns.ui('PhoneField').uri] =
                           var obj = kb.any(subject, property, undefined, store)
                           if (!obj) {
                             obj = kb.any(form, ui('default'))
-                            if (obj) kb.add(subject, property, obj, store)
                           }
                           if (obj && obj.uri && params.uriPrefix) { // eg tel: or mailto:
                             field.value = decodeURIComponent(obj.uri.replace(params.uriPrefix, '')) // should have no spaces but in case


Vinnl:
Thanks @TallTed, I've added Tim as a reviewer now.

Vinnl:
@TallTed It's even "don't delete a field's default value from their Pod" - because that wouldn't work, because it's not _in_ their Pod. The functionality you described would have to be implemented by the form that's using this field.

(I only thought of that today; it was late for me yesterday. Sorry!)

TallTed:
@Vinnl - It's not too late (for you; I don't have the necessary permissions on these) to change the title/summary of this PR (#109), nor of the associated issue (#105) to make things clearer for future readers.

Vinnl:
@TallTed The title is actually correct, I think: because the default value was stored in the rdflib _Store_ (i.e. in-memory), it would then try to delete it from the _Pod_. So this PR stops it from adding it to the Store, which then results in it not trying to delete it from the Pod. (And the issue title identifies the bug that resulted from that behaviour.)

Does that make sense? It's still nice if future readers can see this clarification in the PR comments :)

TallTed:
I suggest changing the title *because this is the first, and often only, thing people read* when reviewing past issues/PRs.  Reading over all the comments requires much more effort -- and is only likely to happen if the title suggests that it's about what they're researching.

Title is now "Don't store a field's default value in the store"

I suggested "don't store this field's default value in the store"
or "don't store this app's user preference default value(s) in the store"

You replied with "don't delete a field's default value from their Pod" which does seem most accurate thus far.

Maybe "apps shouldn't try to save/delete a default field value to/from the user's pod/store" would be even better?

Vinnl:
@TallTed Oh, sorry - lost the context of your previous comment, which actually included a suggestion. Too many update notifications! I'll change it right away.

PR #110, opened 2019-08-28
Adds optional cancelButton to end of tabs

megoth:
By providing option onClose, tabs are extended with an extra, unstyled tab that contains a cancelButton. Once clicked it will call the given onClose-method.

This is needed for providing the cancelButton for the global dashboard.

Part of fix for https://github.com/solid/solid-panes/issues/154

### DIFF
--- a/src/tabs.js
+++ b/src/tabs.js
@@ -26,6 +26,7 @@ const utils = require('./utils')
 // options.showMain(div, subject) function to show subject in div when tab selected
 // options.renderTabSettings  function(subject, domContainer)
 // options.renderTabSettings  like showMain but when user has held Alt down
+// options.onClose            if given, will present a cancelButton next to tabs that calls this optional method
 //
 
 UI.tabs.tabWidget = function (options) {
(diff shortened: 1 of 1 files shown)


megoth:
Rebased onto master, and tested with the test-files.

I did check that the original tests went through, but the test-setup is waaaaaay to slow that I want to use it to test this new part as well.

megoth:
@timbl should I add a new tests, or is it ok to just check the existing tests?

timbl:
The existing tests build a tabs widget and morph its state, with each orientation.  The idea is you can for example check the positioning CSS works. I suggest just adding the option for the X box to the test.   Add a dummy onClose function to the core of the test loop. (Didn't I do that?) so that you can see that the positioning CSS works for that too in each orientation.  I don't think you need to check  that it works without the X button again.
Sorry I thought that was clear

timbl:
Did you run the tests on #110 with the X button included?
onClose() included?
Did the x pup up in a reasonable place for each orientation?
If so I am fine to merge

megoth:
Ok, I've made it kinda work - but we need to have a discussion on the tests; They are not easy to read, and since there are no documentation I have to go a lot back and forth between the HTML, Turtle and JavaScript to make sense of it. I'm _not_ a fan of these kinds of custom test setups, and we should investigate the possibility of making use of snapshots, like [Snapshot Testing in Jest](https://jestjs.io/docs/en/snapshot-testing).

PR #112, opened 2019-08-29
Added methods to filter panes based on user roles (Developer/PowerUser)

megoth:
Adding methods `filterAvailablePanes` (returns which panes a user should see based on their roles) and `getUserRoles` (returns roles for a user).

This is part of fixing https://github.com/solid/solid-panes/issues/167

### DIFF
--- a/src/signin.js
+++ b/src/signin.js
@@ -1306,9 +1306,9 @@ async function getUserRoles () {
 async function filterAvailablePanes (panes) {
   const userRoles = await getUserRoles()
   return Object.values(panes).filter(pane => isMatchingAudience(pane, userRoles))
+}
 
-  function isMatchingAudience (pane, userRoles) {
-    const audience = pane.audience || []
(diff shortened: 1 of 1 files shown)


src/signin.js
```diff
@@ -1287,3 +1289,26 @@ function newAppInstance (dom, appDetails, callback) {
...
+  }
+  await UI.store.fetcher.load(preferencesFile)
+  return UI.store.each(profile, ns.rdf('type'), null, preferencesFile.doc())
+}
+
+async function filterAvailablePanes (panes) {
+  const userRoles = await getUserRoles()
+  return Object.values(panes).filter(pane => isMatchingAudience(pane, userRoles))
+
+  function isMatchingAudience (pane, userRoles) {
```
Vinnl: Somewhat of a nitpick, but I think it'd both be clearer readability-wise and better in terms of memory allocation to not have the definition of `isMatchingAudience` enclosed in `filterAvailablePanes`.

PR #119, opened 2019-09-16
Extracted findProfileImage from findImage in setImage

megoth:
findImage used to be a private function in UI.widgets.buttons.setImage, but the logic can be reused other places, so I've tried to extract the logic so that it becomes a method exposed as UI.authn.findProfileImage.

Partially fixes https://github.com/solid/solid-panes/issues/169

### DIFF
--- a/src/signin.js
+++ b/src/signin.js
@@ -37,7 +37,6 @@ module.exports = {
   filterAvailablePanes, // Async
   findAppInstances,
   findOriginOwner,
-  findProfileImage, // Sync
   getUserRoles, // Async
   loadTypeIndexes,
   logIn,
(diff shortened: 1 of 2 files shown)


src/signin.js
```diff
@@ -1303,6 +1304,19 @@ async function getUserRoles () {
   return UI.store.each(profile, ns.rdf('type'), null, preferencesFile.doc())
 }
 
+function findProfileImage (profile) {
```
timbl: I think moving ths code from buttons.setImage wasa n odd move.  The setImage function in more general and lower level. It should be called by setProfileImage which is specific to a profile image. The setImage can be used by other things as well as profile image — images of people or organizations, and we can add to it more image predicates with time like from schema.org

megoth: My thinking was that findProfileImage belongs to other user-related functions - where would you move it? Or would you make this function reusable in another way?

megoth: Me and @timbl had a chat outside of this - we concluded that we will move findProfileImage back to buttons, and rename it findImage (since it's a more general thing than just connected to profiles).

We also concluded to create https://github.com/solid/solid-ui/issues/120 since we at some point want to clean up the API a bit to make it more intuitive.

megoth:
@timbl I've updated the API - please check that it looks OK.

PR #122, opened 2019-09-17
Do not add an "Add" button when modifying

### LINKED ISSUE #121: Multiple Green Pluses on Sharing Pane
On the Sharing pane, it you continually switch between `Set the sharing of folder contets 
separately from the sharing for the folder` and `Set default for folder contents to
just track the sharing for the folder`, with each click of the first button an additional green plus is displayed on the pane.

## To Replicate
1. Display the Sharing Pane.
2. Click the `Set the sharing of folder contets separately from the sharing for the folder` button.
3. Click the `Set default for folder contents to just track the sharing for the folder` button.

## Expected Behaviour
1. Only one green plus button is displayed.

## Actual Behaviour
1. Two green plus buttons are displayed.

Vinnl:
I really have no idea why these lines of code are there, but given
that they were introduced in a commit that said it was "basically"
working (2e63674702f6c08cef57e771c06f87dbe0a1a0c8), I hope @timbl  can
clarify that.

Since we haven't set up test tooling in solid-ui yet, this PR does not include tests.

Fixes #121.

### DIFF
--- a/src/acl-control.js
+++ b/src/acl-control.js
@@ -526,7 +526,7 @@ UI.aclControl.ACLControlBox5 = function (subject, dom, noun, kb, callback) {
       } // if
     } // for
 
-    if (options.modify) {
+    if (options.modify && !options.doingDefaults) {
       renderAddToolBar(box, lastRow)
     }
 


Vinnl:
Ehm, I guess you're right - let me take another look.

Vinnl:
@megoth Alright, perhaps this does what it's supposed to - can't really be sure though :/

Note that I updated the commit, so if you want to check it out locally, delete the old branch first:

    $ git checkout master 
    $ git branch -D 171-double-plus-ungood
    $ git checkout 171-double-plus-ungood

timbl:
I was expected the solution to the replicating pluses problem to be to move the plus from an outer dom level to an inner one where it vanish with as control is refreshed. ...

timbl:
The title seems reversed.  You only want a green plus when you are modifying the ACL. Not when it is read-only.

megoth:
Closed in favor of https://github.com/solid/solid-ui/pull/182

PR #123, opened 2019-09-17
Look for `doingDefaults` rather than `doingDefault`

Vinnl:
I'm not sure what "default" is supposed to mean, but stepping
through in the debugger showed that the property had an `-s`
appended, I presume that's what it should be?

The code has been there for three years though, so not sure
whether it's even supposed to be doing something?

(Also a +1 for #69, which would have caught this.)

Since solid-ui has no test tooling set up yet, this PR does not include tests.


Vinnl:
@timbl Alright, done.

PR #129, opened 2019-09-24
Do not allow : or # to be part of name for resource

megoth:
After testing out reserved characters in URL (https://developers.google.com/maps/documentation/urls/url-encoding), I also found that `#` is problematic to allow in names for resources. So I've added functionality that disallows creating resources with name `#` and `:`.

Fix https://github.com/solid/solid-panes/issues/178

### DIFF
--- a/src/create.js
+++ b/src/create.js
@@ -202,7 +202,7 @@ function newThingUI (context, thePanes) {
 }
 
 function validatePathname (pathname) {
-  return !(/[#:]/g).test(pathname)
+  return pathname.indexOf('#') === -1 && pathname.indexOf(':') === -1
 }
 
 // Form to get the name of a new thing before we create it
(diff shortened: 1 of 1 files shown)


src/create.js
```diff
@@ -197,6 +201,10 @@ function newThingUI (context, thePanes) {
   }
 }
 
+function validatePathname (pathname) {
+  return !(/[#:]/g).test(pathname)
```
Vinnl: I'd avoid regular expressions if possible, since they're relatively slow (not really a problem outside of hot code paths, such as this) and often hard to read. In this case, I'd just go for two `indexOf`'s.
```suggestion
  return pathname.indexOf('#') === -1 && pathname.indexOf(':') === -1
```

angelo-v:
@Vinnl @megoth @timbl Can you tell why the progress here stopped? Should it be resolved and merged, or closed?

Vinnl:
I don't know, unfortunately. @megoth?

megoth:
@angelo-v I can't remember why we stopped progress on this either =/

angelo-v:
Do you think it is still needed? Does not appear so to me, see https://github.com/solid/solid-panes/issues/178#issuecomment-734955559

megoth:
I honestly cannot remember why we needed it originally, and nothing comes to mind from bugs I've recently encountered, so I think it's safe to close.

PR #144, opened 2019-10-10
Fix problem when dropping .mp3 files getting double suffixes

timbl:
To address https://github.com/solid/folder-pane/issues/5


bourgeoa:
see comments https://github.com/solid/folder-pane/issues/5#issuecomment-542262762

PR #147, opened 2019-10-15
delete check contentType match file extension

bourgeoa:
checked done in NSS resource-mapper.mapUrlToFile


timbl:
Why remove the check?    The databrowser world works by content types and extensions being in sync.

timbl:
See https://github.com/solid/solid-ui/pull/144 which fixes a problem with this function

bourgeoa:
See PR #156. I am unable to resolve conflicts

PR #150, opened 2019-10-23
contentType mismatch solved by NSS

bourgeoa:
Related: https://github.com/solid/node-solid-server/pull/1328

Fixes https://github.com/solid/folder-pane/issues/5

should replace PR #144 that should be closed


bourgeoa:
unable to solve conflicts

PR #151, opened 2019-10-25
Include ts files in babel builds

fletcher91:
This fixes an issue created by https://github.com/solid/solid-ui/pull/127, where the index (the only typescript file) is not compiled to javascript since babel doesn't recognize its file extension.

The missing `lib/index.js` file causes `module not found` exceptions in downstream projects

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,13 +10,13 @@
   ],
   "scripts": {
     "build": "npm run build-lib && npm run build-types",
-    "build-lib": "babel src -d lib --source-maps",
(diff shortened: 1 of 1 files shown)


fletcher91:
This is causing quite a lot of downstream issues in my current setup. The original changes (127) moves to ES modules which export a `default` object in commonjs, causing errors in solid-panes.

E.g. `UI.icons.originalIconBase` doesn't work anymore, but `UI.icons.default.originalIconBase` does.

PR #153, opened 2019-11-07
Reintroducing use of byCombo

megoth:
Should fix https://github.com/solid/solid-panes/issues/193


megoth:
I cannot find any use of `doingDefaults` outside of solid-ui and no use of `doingDefault` at all (and I've checked all projects connected to the data browser). But yes, there might be other projects that use this API that uses `doingDefault`, so we'll double-check with @timbl .

Vinnl:
I wasn't really talking about other projects, just that the code you just put back used to refer to `doingDefault` (which didn't exist) - so whatever that code was supposed to do, it never did it, but it will likely now start doing it. I have no idea about whether that's desirable or not.

PR #154, opened 2019-11-08
Adding caching of user roles

megoth:
We do not want to spam profile and settings every time we navigate to another point in file structure (e.g. using the Folder view)

### DIFF
--- a/src/signin.js
+++ b/src/signin.js
@@ -27,6 +27,8 @@ const UI = {
   widgets: require('./widgets') // 2018-07-31
 }
 
+const cachedResponses = {}
+
 const ns = UI.ns
 const kb = UI.store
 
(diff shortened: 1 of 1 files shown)


megoth:
There is no explicit issue for this, but it was inspired by my research when investigating https://github.com/solid/folder-pane/issues/6.

timbl:
What do you mean "spam"?  Spam means sending useless information to.  Do you mean "making unnecessary reads" of?   But the profile and preferences are already cached in the store.  Do you mean "avoiding unnecessary queries of the triple store"?   Is that a problem?

megoth:
Ah yes, the use of the term spam is inaccurate - I meant it in the sense that there might be unnecessary amount of reads.

I can't remember exactly why I though this was necessary (it's one month old), but I might've simply forgotten that profile and preferences are already cached. Since that is the case, this is an unnecessary change, and I'll close this PR.

timbl:
The cache you have there is global, and has no mechanism to be invalidated when the logged in user changes.  I suggest you add the user roles to the existing login context, which I guess in now part of the databrowser context?    And make suer that when the user logs out is is dropped.    scrapped on any non-public da
(Hmmm -- when the user logs out should the quadstore in fact be stripped of any non-public data?)

PR #166, opened 2019-11-28
Migrating ACL module to TypeScript

megoth:
Showcasing how migration can be done.

Note that I had to do cast a lot of types because of previously untyped variables that can be various types. We should avoid these ambiguous variables in the future, as it becomes difficult to read and can lead to some surprising results.

I also had to throw a HTMLElement to any (line 91 in aclControl.ts) as code later on heavily rely on the ability to add properties to this variable. This is also something we should avoid, as using undocumented properties might open up for bugs later on.


megoth:
An important part of the migration is to add types to all public methods, both for parameters and what they are expected to return.

src/acl/acl-control.ts
```diff
@@ -49,87 +45,100 @@ UI.aclControl.preventBrowserDropEvents = function (document) {
...
+    box,
+    doc,
+    aclDoc,
+    kb,
+    options: {
+      doingDefaults?: boolean,
+      modify?: boolean
+    } = {}) {
+    const ac = readACL(doc, aclDoc, kb, options.doingDefaults) // Note kb might not be normal one
+    const byCombo = ACLbyCombination(ac) // important to keep this, as removal will trigger bugs like https://github.com/solid/solid-panes/issues/193#issuecomment-549777549
```
timbl: Don't need "important to keep" notice .. no real reason for people noit to keep it.  It was removed once because of a misunderstanding.  (https://github.com/solid/solid-ui/commit/5afb141378f195f395f249258a7fd02768334ada#commitcomment-36205141 )

src/acl/acl-control.ts
```diff
@@ -148,21 +157,21 @@ UI.aclControl.ACLControlBox5 = function (subject, context, noun, kb, _callback)
...
           list += y[i]
         }
       }
       return list
     }
 
-    var removeAgentFromCombos = function (uri) {
-      for (var k = 0; k < 16; k++) {
-        var a = byCombo[kToCombo(k)]
+    const removeAgentFromCombos = function (uri) {
```
timbl: Use  normal function  syntax instead of const?

src/acl/acl-control.ts
```diff
@@ -435,27 +437,26 @@ UI.aclControl.ACLControlBox5 = function (subject, context, noun, kb, _callback)
...
-        UI.widgets.makeDropTarget(row, handleManyDroppedURIs)
+        widgets.makeDropTarget(row, handleManyDroppedURIs)
       }
       return row
     } // renderCombo
 
-    var syncPanel = function () {
-      var kids = box.children
-      for (var i = 0; i < kids.length; i++) {
+    const syncPanel = function () {
```
timbl: function syntax?

src/acl/acl-control.ts
```diff
@@ -710,31 +712,24 @@ UI.aclControl.ACLControlBox5 = function (subject, context, noun, kb, _callback)
     return byCombo
   } // ACLControlEditable
 
-  var renderBox = function () {
+  const renderBox = function () {
```
timbl: function syntax

megoth:
@timbl Have implemented your requested changes ^_^

PR #177, opened 2019-12-19
Refactoring to improve readability

megoth:
This is huge refactor (restructuring code without changing external behavior).

With this I hope to improve readability of the code, to make it easier for newcomers to come in to the code and be able to change code without fear of not understanding the code properly.

Note that there is still improvements to be done wrt some of the uses of scopes (they are very deep at times, and state is difficult to follow), but I didn't want to risk breaking any code by changing this behavior to much.

This is also an example of how _I think_ code can be improved in general, and is an on-going discussion with Tim and I.


megoth:
Note: The changes are so many that it probably makes more sense to look at the new file in itself and just review the code structure. As this is a refactor the exposed functionality should not have been affected.

megoth:
I think I have another way to improve this code - will close this PR for now.

PR #179, opened 2019-12-27
Deal with empty tools list

michielbdejong:


michielbdejong:
In the end it turns out the tool list was empty because of https://github.com/solid/meeting-pane/pull/9#issuecomment-570537533

PR #180, opened 2019-12-27
Icons potentially for Pod, Class, Property

timbl:
from the usual thenounproject.com


timbl:
Screen Shot 2019-12-31 at 15 32 36
Where we have an icon for a thing of a particular class, then we should use that icon in a green square or rectangle for the icon of the class itself

PR #182, opened 2020-01-09
Restructured ACLControlBox5 into multiple modules

megoth:
Is now split into AccessController, AccessGroups, and AddAgentButtons. All methods that are related to DOM manipulation are prefixed `render` - the rest deals with data manipulation and state handling. Further, no non-`render`-methods should call on any `render`-methods.

This (I think) makes the code easier to read and maintain.

Have also added JSS to showcase how dynamic CSS can work. (Is not much work to remove if not wanted.)

Fixes https://github.com/solid/solid-ui/issues/121.

### DIFF
--- a/src/acl/access-controller.ts
+++ b/src/acl/access-controller.ts
@@ -36,14 +36,6 @@ export class AccessController {
     this.isContainer = targetDoc.uri.slice(-1) === '/' // Give default for all directories
     if (defaultHolder && defaultACLDoc) {
       this.isUsingDefaults = true
-      const defaults = this.store
-        .each(null, ACL('default'), defaultHolder, defaultACLDoc)
-        .concat(this.store.each(null, ACL('defaultForNew'), defaultHolder, defaultACLDoc))
-      if (!defaults.length) {
-        this.status.innerText = ' (No defaults given.)'
-      } else {
-        this.status.innerText = ''
-      }
       const aclDefaultStore = adoptACLDefault(this.targetDoc, targetACLDoc, defaultHolder, defaultACLDoc)
       this.mainCombo = new AccessGroups(targetDoc, targetACLDoc, this, aclDefaultStore, { defaults: true })
       this.defaultsCombo = null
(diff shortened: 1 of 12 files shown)


src/acl/access-groups.ts
```diff
@@ -0,0 +1,274 @@
...
+  public render (): HTMLElement {
+    this.root.innerHTML = ''
+    this.renderGroups().forEach(group => this.root.appendChild(group))
+    if (this.controller.isEditable) {
+      this.root.appendChild(this.addAgentButton.render())
+    }
+    return this.root
+  }
+
+  private renderGroups (): HTMLElement[] {
```
timbl: Let's leave the word 'render' for things which make HTML DOM.    Suggest  getGroups here

src/acl/access-groups.ts
```diff
@@ -0,0 +1,274 @@
...
+    widgets.makeDropTarget(groupRow, (uris) => this.handleDroppedUris(uris, combo)
+      .then(() => this.controller.render())
+      .catch(error => this.controller.renderStatus(error)))
+    const groupColumns = this.renderGroupElements(comboIndex, combo)
+    groupColumns.forEach(column => groupRow.appendChild(column))
+    return groupRow
+  }
+
+  private renderGroupElements (comboIndex, combo): HTMLElement[] {
+    const groupName = this.controller.dom.createElement('div')
```
timbl: Variable naming - this confused me.  groupName I would expect to be the name of a group, wither a string or an RDF literal.  Prefer to call an element an element  ...   like groupNameDiv or groupNameElement or groupNameRow etc  or maybe groupNameEle

src/acl/access-groups.ts
```diff
@@ -0,0 +1,274 @@
...
+  private async deleteAgent (combo, pred, obj): Promise<void> {
+    const combos = this.byCombo[combo] || []
+    const comboToRemove = combos.find(([comboPred, comboObj]) => comboPred === pred && comboObj === obj)
+    if (comboToRemove) {
+      combos.splice(combos.indexOf(comboToRemove), 1)
+    }
+    await this.controller.save()
+  }
+
+  public updateStore (store: IndexedFormula): void {
```
timbl: Name of function - 'loadAndConvertACL' ?    It does load the ACL into the store, but also mainly into the internal byCombo form in this object.   Prefer to use 'update' for changes going back to the pod, 'load' for things coming in the read direction.

src/acl/access-groups.ts
timbl: Oh I see it does return DOM elements, it was the names of those  variables were confusing.

src/acl/access-groups.ts
megoth: Fair enough, and I think that's a good convention. My latest commit addresses this.

src/acl/access-groups.ts
megoth: I realized this might be better to fix with a setter. I've committed a solution using this approach in the latest commit.

src/acl/access-controller.ts
```diff
@@ -0,0 +1,242 @@
...
+      this.rootElement.appendChild(this.renderAddAclsController())
+    } else if (!this.targetIsProtected) {
+      this.rootElement.appendChild(this.renderRemoveAclsController())
+    }
+    return this.rootElement
+  }
+
+  private renderRemoveAclsController (): HTMLElement {
+    const useDefaultButton = this.dom.createElement('button')
+    useDefaultButton.innerText = `Stop specific sharing for this ${this.noun} -- just use default${this.prospectiveDefaultHolder ? ` for ${utils.label(this.prospectiveDefaultHolder)}` : ''}`
```
TallTed:
```suggestion
    useDefaultButton.innerText = `Remove custom permissions for this ${this.noun} -- just use default${this.prospectiveDefaultHolder ? ` for ${utils.label(this.prospectiveDefaultHolder)}` : ''}`
```

PR #189, opened 2020-02-03
Basic jest and typedoc

michielbdejong:


michielbdejong:
Continued in #191

michielbdejong:
Reopening this so Vincent can review my changes separately from Sharon's ongoing work on tests for  `src/pad.js`, `src/widgets/forms.js`, and `src/tabs.js`.

package.json
```diff
@@ -12,12 +12,16 @@
...
+    "build-version": "sh ./timestamp.sh > src/versionInfo.ts  && eslint \"src/versionInfo.ts\" --fix",
     "watch": "npm run build-version && babel src --out-dir lib --watch --source-maps --extensions '.ts,.js'",
     "clean": "rm -rf lib",
-    "lint": "eslint 'src/**/*.js' 'src/**/*.ts'",
-    "lint-fix": "eslint 'src/**/*.js' 'src/**/*.ts' --fix",
-    "test": "npm run lint",
+    "coverage": "jest --coverage --collectCoverageFrom=src/**/*.[jt]s",
+    "lint": "eslint \"src/**/*.js\" \"src/**/*.ts\"",
+    "lint-fix": "eslint \"src/**/*.js\" \"src/**/*.ts\" --fix",
+    "jest": "jest",
```
Vinnl: I don't think this really adds anything?

src/chat/dateFolder.js
```diff
@@ -93,7 +93,7 @@ module.exports = class DateFolder {
     const folder = this.leafDocumentFromDate(date).dir()
     const found = await previousPeriod(folder, 3)
     if (found) {
-      const doc = kb.sym(found.uri + this.leafFileNam)
+      const doc = kb.sym(found.uri + this.leafFileName)
```
Vinnl: So, ehm... This never worked? :P

test/helpers/dom.ts
```diff
@@ -0,0 +1,37 @@
+export const element = {
```
Vinnl: All this should already be provided by jsdom (`document.createElement()`), with actual mock implementations?

test/unit/acl/access-controller.test.ts
```diff
@@ -0,0 +1,31 @@
+jest.mock('rdflib')
+// import * as RdfLib from 'rdflib'
+jest.mock('solid-auth-client')
+// import * as SolidAuthClient from 'solid-auth-client'
+// import { DataBrowserContext } from 'pane-registry'
```
Vinnl: Uncommented lines can be removed.

test/unit/acl/access-controller.test.ts
```diff
@@ -0,0 +1,31 @@
...
+const AccessController = require('../../../src/acl/access-controller')
+// just a note that the only public functions are isEditable, render, and save
+describe('AccessController', () => {
+  debugger
```
Vinnl:
```suggestion
```

test/unit/acl/access-controller.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  it('exists', () => {
+    expect(AccessController).toBeTruthy()
+  })
+  /* leaving this for now
```
Vinnl: Maybe specify why it's left in so future readers know whether they can remove it.

test/unit/acl/access-groups.test.ts
```diff
@@ -0,0 +1,12 @@
+jest.mock('rdflib')
+import * as RdfLib from 'rdflib'
+jest.mock('solid-auth-client')
+import * as SolidAuthClient from 'solid-auth-client'
+
+import * as AccessGroups from '../../../src/acl/access-groups'
+
+describe('AccessGroups', () => {
+  it('exists', () => {
+    expect(AccessGroups).toBeTruthy()
```
Vinnl: There is no way this could ever not be truthy - if you call `import * as AccessGroups`, then even if it does not export anything you would get `{}`. (Likewise in other tests.)

test/unit/acl/access-groups.test.ts
```diff
@@ -0,0 +1,12 @@
+jest.mock('rdflib')
+import * as RdfLib from 'rdflib'
+jest.mock('solid-auth-client')
+import * as SolidAuthClient from 'solid-auth-client'
```
Vinnl: These imports are unused? (Likewise in other tests.)

michielbdejong: You need to import them for jest.mock to work, right?

Vinnl: No, `jest.mock` makes sure that whenever the code under test imports the given module, that it will receive the mocked replacement. If you import then yourself, I think that just means you have an instance of the mocks in your test file, but you're not using them anywhere.

package.json
michielbdejong: It's needed if you want to run only the jest tests. I tried removing it but then I see:
```sh
~/gh/solid/solid-ui $ npm run jest
npm ERR! missing script: jest
npm ERR! 
npm ERR! Did you mean this?
npm ERR!     test

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/michiel/.npm/_logs/2020-02-10T19_14_04_093Z-debug.log

File: test/helpers/dom.ts (hunk 3 above)
Comment (michielbdejong) [reply test/helpers/dom.ts]: Yes, `document.createElement` is a dependency of solid-ui that is available in the browser as well as (via jsdom) in node. The same is true for rdflib and solid-auth-client, but I'm also mocking those, because none of them are part of the system under test, right? We discussed this on 6 Feb, and then we agreed to [mock anything that produces side effects](https://inrupt-team.slack.com/archives/CQKEE9G3W/p1580990593020800). Do you think we should revise that decision?

Comment (michielbdejong) [reply test/helpers/dom.ts]: Maybe in the unit tests we mock it, but in the integration tests we use jsdom?

File: test/unit/acl/access-controller.test.ts (hunk 4 above)
Comment (michielbdejong) [reply test/unit/acl/access-controller.test.ts]: Ah sorry, fixed.

File: test/unit/acl/access-controller.test.ts (hunk 6 above)
Comment (michielbdejong) [reply test/unit/acl/access-controller.test.ts]: Ah sorry, fixed

File: test/unit/acl/access-controller.test.ts (hunk 5 above)
Comment (michielbdejong) [reply test/unit/acl/access-controller.test.ts]: Ah sorry, fixed

File: test/unit/acl/access-groups.test.ts (hunk 7 above)
Comment (michielbdejong) [reply test/unit/acl/access-groups.test.ts]: Right, good point, I'll fix those tomorrow.

File: test/unit/acl/access-groups.test.ts (hunk 8 above)
Comment (michielbdejong) [reply test/unit/acl/access-groups.test.ts]: Ah ok! Good to know, I'll remove them tomorrow.

File: test/helpers/dom.ts (hunk 3 above)
Comment (Vinnl) [reply test/helpers/dom.ts]: JSDOM already _is_ a mock of the DOM :) You should generally be able to assume that JSDOM will not have bugs - and the same holds true for rdflib and solid-auth-client. But you mock those because they make HTTP requests and you don't want to rely on some server being up and serving the right requests, and because you want to influence what they return. So you might want s-a-c to tell the system under test that a given WebID is logged in that you then make assertions on, or you instead that the user is logged out.

In the case of DOM elements, however, there is going to be no difference in the implementation - you always want `document.createElement` to return an element that e.g. has an `appendChild`.

File: package.json (hunk 1 above)
Comment (Vinnl) [reply package.json]: Try `npx jest`.

File: test/unit/acl/access-groups.test.ts (hunk 8 above)
Comment (michielbdejong) [reply test/unit/acl/access-groups.test.ts]: Done

File: test/unit/acl/access-groups.test.ts (hunk 7 above)
Comment (michielbdejong) [reply test/unit/acl/access-groups.test.ts]: Done

File: test/helpers/dom.ts (hunk 3 above)
Comment (michielbdejong) [reply test/helpers/dom.ts]: Using jsdom now.

File: test/unit/acl/access-controller.test.ts [hunk 9]
```diff
@@ -0,0 +1,89 @@
+import * as RdfLib from 'rdflib'
```
Vinnl: Hmm, maybe make sure that you actually get rdflib here, rather than the mocks (Jest hoists `jest.mock` calls to the top, I believe). Otherwise that will probably be a surprise in the future.

Otherwise, you could look at using `require.requireActual` to get the unmocked version.

michielbdejong: We never want real rdflib or real s-a-c to be used, right?

michielbdejong: as discussed :)

PR #195, opened 2020-02-12
Small ACL cleanup

michielbdejong:
Will try to get branch coverage and API documentation of `src/acl/` to 100%.
Starting from 13% for the folder average (with a 0 for `add-agent-buttons.ts`, somehow?):
Screenshot 2020-02-12 11 32 26

### DIFF
--- a/__mocks__/rdflib.ts
+++ b/__mocks__/rdflib.ts
@@ -41,40 +41,76 @@ export function variable () {
 }
 
 export function graph() {
-  return {
-    any: () => {
-      return sym()
-    },
-    each: () => {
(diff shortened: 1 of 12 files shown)


michielbdejong:
OK, one issue with getting branch coverage for this is that the module produces a button, with a lot of code behind it that will only work when you click it. And I think that some of that code that runs when you click is even adding more code dynamically. I'll ponder a bit how I can write good tests for all that.

michielbdejong:
Didn't do very much, but at least dipped my toe into button clicking tests. :)

PR #201, opened 2020-02-14
Old gh-pages branch

michielbdejong:
Just to make sure we don't lose any old work, there was a gh-pages branch from 2017; I renamed it to `gh-pages-old`. This PR contains that changes that that branch had, but master doesn't.

We can close this PR if we are confident we don't need to rebase any of those old changes onto current master.


angelo-v:
gh-pages works fine

PR #204, opened 2020-02-14
Buttons

michielbdejong:


src/widgets/buttons.ts
```diff
@@ -184,19 +189,37 @@ function setName (element, x) {
...
+ *          * ns.foaf('depiction')
+
+ */
+function imagesOf (x: NamedNode, kb: IndexedFormula): any[] {
   var ns = UI.ns
   return kb
     .each(x, ns.sioc('avatar'))
     .concat(kb.each(x, ns.foaf('img')))
     .concat(kb.each(x, ns.vcard('logo')))
+    .concat(kb.each(x, ns.vcard('hasPhoto')))
```
michielbdejong: To align with `findImage`

michielbdejong:
Screenshot 2020-02-19 21 20 44

michielbdejong:
@SharonStrats let's solve the webpack build problem in a separate issue:
https://github.com/solid/solid-ui/issues/213
In the meantime you can check the examples page at: http://solid.github.io/solid-ui/examples/buttons/
Please re-review

src/widgets/buttons.ts
```diff
@@ -24,18 +26,25 @@ const dragAndDrop = require('./dragAndDrop')
...
+  let dom = context && context.dom
   if (!dom && typeof document !== 'undefined') {
     dom = document
   }
   if (dom) {
     var body = dom.getElementsByTagName('body')[0]
-    box = dom.createEvent('div')
-    body.insertBefore(box, body.firstElementChild)
-    context.statusArea = box
+    ;(box as unknown as HTMLElement) = dom.createElement('div')
```
SharonStrats: Having to use 'as' will be something we look to change when we refactor to Typescript completely?

src/widgets/buttons.ts
```diff
@@ -332,10 +373,10 @@ function trySetImage (element, thing, iconForClassMap) {
   return false // we can do better
 }
 
-// ToDo: Also add icons for *properties* like  home, work, email, range, domain, comment,
-//
-
-function setImage (element, thing) { // 20191230a
+/**
```
SharonStrats: We would like to change // to /*  */?

src/widgets/buttons.ts
michielbdejong: Ah actually there the cast was unnecessary.

In other places I cast from `HTMLElement | null` to `HTMLElement` to indicate that the code ignores the possibility of `document.createElement('div')` returning null.

PR #206, opened 2020-02-16
Notepad copy

timbl:
As a notepad user I want to be able to get/export an HTML copy to use elsewhere.

### DIFF
--- a/src/authn/authn.ts
+++ b/src/authn/authn.ts
@@ -315,9 +315,9 @@ async function loadIndex (
 ): Promise<AuthenticationContext> {
   // Loading preferences is more than loading profile
   try {
-    ;(await isPublic)
+    context = await (isPublic
       ? logInLoadProfile(context)
-      : logInLoadPreferences(context)
+      : logInLoadPreferences(context))
   } catch (err) {
     widgets.complain(context, `loadPubicIndex: login and load problem ${err}`)
   }
--- a/src/chat/infinite.js
+++ b/src/chat/infinite.js
@@ -314,7 +314,7 @@ export function infiniteMessageArea (dom, kb, chatChannel, options) {
         UI.media.cameraButton(dom, kb, getImageDoc, tookPicture)
       )
 
-      UI.pad.recordParticipation(chatChannel, chatChannel.doc()) // participation =
+      UI.participation.recordParticipation(chatChannel, chatChannel.doc()) // participation =
     } // turn on inpuut
 
     const context = { div: middle, dom: dom }
--- a/src/index.ts
+++ b/src/index.ts
@@ -59,7 +59,7 @@ import { infiniteMessageArea } from './chat/infinite'
 // @ts-ignore
 import pad from './pad'
 // @ts-ignore
-import participation from './participation'
+import participation from './participation' // 20200229
 // @ts-ignore
 import preferences from './preferences'
 // @ts-ignore
--- a/src/index.ts
+++ b/src/index.ts
@@ -59,6 +59,8 @@ import { infiniteMessageArea } from './chat/infinite'
 // @ts-ignore
 import pad from './pad'
 // @ts-ignore
+import participation from './participation'
+// @ts-ignore
 import preferences from './preferences'
 // @ts-ignore
 import store from './store'
@@ -91,6 +93,7 @@ if (typeof window !== 'undefined') {
     messageArea,
     infiniteMessageArea,
     pad,
+    participation,
     preferences,
     store,
     style,
--- a/src/index.ts
+++ b/src/index.ts
@@ -120,6 +120,7 @@ export {
   messageArea,
   infiniteMessageArea,
   pad,
+  participation,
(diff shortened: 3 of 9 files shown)


src/authn/authn.ts
```diff
@@ -315,9 +315,9 @@ async function loadIndex (
 ): Promise<AuthenticationContext> {
   // Loading preferences is more than loading profile
   try {
-    ;(await isPublic)
+    context = await (isPublic
```
michielbdejong: Why overwrite `context` which was passed in from https://github.com/solid/solid-ui/blob/master/src/authn/authn.ts#L312?
And also from https://github.com/solid/solid-ui/blob/master/src/authn/authn.ts#L314 it seems that `isPublic` is a boolean?

I'll dig into this function later today, adding unit tests, code documentation, and a usage example, so we can all understand better what the desired behaviour is here. Without that, this is very hard for me to review. Will review the rest of this PR independently first.

src/chat/infinite.js
```diff
@@ -314,7 +314,7 @@ export function infiniteMessageArea (dom, kb, chatChannel, options) {
         UI.media.cameraButton(dom, kb, getImageDoc, tookPicture)
       )
 
-      UI.pad.recordParticipation(chatChannel, chatChannel.doc()) // participation =
+      UI.participation.recordParticipation(chatChannel, chatChannel.doc()) // participation =
```
michielbdejong: I'll review this change separately, it seem unrelated to both this PR and this commit.

src/participation.js
```diff
@@ -0,0 +1,197 @@
...
+  } catch (e) {
+    container.appendChild(
+      UI.widgets.errorMessageBlock(
+        dom,
+        'Error recording your partipation: ' + e
+      )
+    ) // Clean up?
+  }
+  return table
+}
```
michielbdejong: This whole commit seems to be about moving participation out of src/pad.js and into its own src/participation.js

In itself that looks like a good thing! But unrelated to notepad copy, so I'll move this into a separate PR. Also, in the new team we established a rule-of-thumb that we move any code we touch to typescript, and i would also like to add docs, unit tests, and examples in the process.

src/pad-utils.js
```diff
@@ -32,36 +32,44 @@ export function xmlEncode (str) {
 export function notePadToHTML (pad, kb) {
   const chunks = getChunks(pad, kb)
   var html = '<html>\n  <head>\n'
-  const title = kb.anyValue(pad, ns.DCT('title'))
+  const title = kb.anyValue(pad, ns.dct('title'))
```
michielbdejong: this change seems unrelated?

src/authn/authn.ts
michielbdejong: Moving this to https://github.com/solid/solid-ui/issues/209

src/chat/infinite.js
michielbdejong: Continued in https://github.com/solid/solid-ui/issues/210

src/index.ts
```diff
@@ -59,6 +59,8 @@ import { infiniteMessageArea } from './chat/infinite'
 // @ts-ignore
 import pad from './pad'
 // @ts-ignore
+import participation from './participation'
+// @ts-ignore
```
michielbdejong: Moved to https://github.com/solid/solid-ui/issues/210

src/pad-utils.js
```diff
@@ -0,0 +1,75 @@
+
+const $rdf = require('rdflib')
+const ns = require('./ns')
+const PAD = $rdf.Namespace('http://www.w3.org/ns/pim/pad#')
+
+/* Get the chunbks of the noepad
```
michielbdejong: Typo

src/pad.js
```diff
@@ -13,15 +15,22 @@ var UI = {
...
   rdf: $rdf,
   store: require('./store'),
   widgets: require('./widgets')
 }
-const kb = UI.store
-const ns = UI.ns
 
 const utils = require('./utils')
 
+// For back-compatiuble API only until next major release: @@ todo: remove these three
```
michielbdejong: typo

src/participation.js
SharonStrats: If you want I can take a look at this, as I was working on pad.js testing?  And there were also some complexity warnings that I had either resolved or was working on.  Maybe I can combine these into one change.

michielbdejong: If you're offering, I won't say no! :) https://github.com/solid/solid-ui/issues/210#event-3056026823

michielbdejong:
@timbl can this be closed now that #211 is merged?

michielbdejong:
See example usage here: http://solid.github.io/solid-ui/examples/notepad/

megoth:
> @timbl can this be closed now that #211 is merged?

I'll await reviewing until this has a response.

angelo-v:
If I understand this correctly, this PR has been superseded.

PR #214, opened 2020-02-23
Widget tests2

SharonStrats:
I'd like to merge this in now, so that I can clean up and start working on another branch.  

-- Main additions are the tests in peoplePicker.test.ts and a few other test files.
-- The two more sensitive changes are to peoplePicker.js which are moving findAddressBook and createNewGroup functions to be standalone and exported.


megoth:
Closing this, as it is duplicated by https://github.com/solid/solid-ui/pull/207.

test/unit/widgets/peoplePicker.test.ts
```diff
@@ -5,21 +5,112 @@ import {
...
-// jest.mock('../../../src/store')
+const kb = require('../../../src/store')
+const fetcher = kb.fetcher
 
 const dom = new JSDOM('<!DOCTYPE html><p>Hello world</p>').window.document
 const element = dom.createElement('div')
 
+describe('FindAddressBook', () => {
+  it('exists', () => {
+    expect(findAddressBook).toBeInstanceOf(Function)
```
megoth: Do not need tests that check that an imported function exists - TypeScript does this for us.

If anything, we might want to write these tests to make sure that the functionality are exposed through the public API, especially for aliases. But as a rule of thumb, you do not need to test for the existence of functionality, since TypeScript handles this for us.

test/unit/widgets/peoplePicker.test.ts
```diff
@@ -53,7 +143,53 @@ describe('PeoplePicker.render', () => {
...
+
+    peoplePicker.render()
+    // expect(spyOnNowOrWhenFetched).toBeCalled()
+    // expect(spyOnNowOrWhenFetched).toReturnWith('book')
+    expect(spyOnNowOrWhenFetched).toBeCalledTimes(1)
+    // expect(spyAny).toBeCalled() // not getting called.
+    // expect(spySecondAny).toBeCalled()
+    // expect(spyLoad).toBeCalled()
+
+    expect(peoplePicker.render()).toMatchInlineSnapshot(`
```
megoth: In general I think we should try to avoid inline snapshots, and have actual snap files using `toMatchSnapshot`.

test/unit/widgets/peoplePicker.test.ts
```diff
@@ -388,3 +590,49 @@ describe('GroupBuilder.setGroupName', () => {
...
+describe('Person', () => {
+  it('exists', () => {
+    expect(new Person()).toBeInstanceOf(Person)
+  })
+})
+describe('Person.render', () => {
+  it('exists', () => {
+    expect(new Person().render).toBeInstanceOf(Function)
+  })
+  it('runs', () => {
```
megoth: Try to be a bit more specific about what the test does - "runs" is a bit to generic.

test/unit/widgets/peoplePicker.test.ts
```diff
@@ -5,21 +5,112 @@ import {
...
+  })
+  it('should throw an error when there an error with the load', () => {
+    // this was working before I changed the above to mockReturnValueOnce
+    // which I needed to do in order to get that test to take affect...
+    // need to research
+    const spyOnNowOrWhenFetched = jest.spyOn(fetcher, 'nowOrWhenFetched')
+    const spyAny = jest
+      .spyOn(kb, 'any')
+      .mockReturnValue('book')
+      .mockReturnValue('book')
```
megoth: Just FYI: When I see this chunks of code that looks very similar, it can be tempting to write some helper functionality that abstracts these chunks. Than again, that might make it harder to modify tests later, so it's a difficult line to balance. For now, lets avoid the temptation and live with the redundant chunks of code.

test/unit/widgets/peoplePicker.test.ts
SharonStrats: Thank you.  I will remove this one.

test/unit/widgets/peoplePicker.test.ts
SharonStrats: I will change this.

test/unit/widgets/peoplePicker.test.ts
michielbdejong: I disagree, see https://github.com/solid/solid-ui/blob/master/src/widgets/index.js#L23.

michielbdejong: Let me try it out and test what @megoth said, please don't remove it for now.

megoth: Again, as a rule of thumb. The example you refer to makes sense to have a unit test that test that the API we export is as expected.

Than again, I would refactor the code that you linked to so that we are explicit about which things we import and export, instead of simply smashing all methods from existing modules into one object using `Object.assign`. Then we would not need unit tests, as TypeScript would ensure that these methods exist.

michielbdejong: Discussion continued in https://github.com/solid/solid-ui/pull/218

michielbdejong: Even if you refactor widgets/index, then widgets/peoplePicker needs its own unit tests.
I'm sure that somewhere in solid-panes there is some code that would break if we would merge #218, but that's not how unit tests work. They should test the unit (widgets/peoplePicker) in isolation without relying on how other files use what it exports.

michielbdejong: So let's *not* remove any of the "exists" tests, please.

megoth: Answered in #218

michielbdejong: :+1:

jest.config.ts
```diff
@@ -1,4 +1,5 @@
 module.exports = {
   verbose: true,
-  prettierPath: 'prettier'
+  prettierPath: 'prettier',
```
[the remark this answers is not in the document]
michielbdejong: Resolved in https://github.com/solid/solid-ui/issues/220, right?
@SharonStrats please confirm if that approach works for what you want to do with snapshots.

.prettierrc
```diff
@@ -1,5 +1,5 @@
 {
-  "arrowParens": "always",
+  "arrowParens": "avoid",
```
megoth: Did you deem Prettier necessary for this to work? `.toMatchSnapshot` did not work as intended?

package.json
```diff
@@ -10,11 +10,11 @@
   ],
   "scripts": {
     "build": "npm run clean && npm run build-version && npm run build-lib && npm run build-types && npm run build-webpack",
-    "build-lib": "babel src -d lib --source-maps --extensions '.ts,.js'",
```
megoth: I suspect none of the changes in this file is necessary? Please revert the changes if that's the correct

src/widgets/peoplePicker.js
```diff
@@ -28,6 +28,7 @@ export class PeoplePicker {
     this.groupPickedCb = groupPickedCb
     this.selectedgroup = this.options.selectedgroup // current selected group if any
     this.onSelectGroup = this.onSelectGroup.bind(this)
+    this.findAddressBook = this.findAddressBook.bind(this)
```
megoth: You don't need to do this if it's already moved into the class. This happens implicitly.

jest.config.ts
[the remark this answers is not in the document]
SharonStrats: Confirmed that regular snapshots work without prettier.  I have changed Inline to just toMatchSnapshot

.prettierrc
SharonStrats: I have changed to toMatchSnapshot

src/widgets/peoplePicker.js
```diff
@@ -43,29 +43,29 @@ export class PeoplePicker {
...
             new GroupPicker(container, book, this.onSelectGroup).render()
           })
 
           const createNewGroupButton = document.createElement('button')
           createNewGroupButton.textContent = escape('Create a new group')
           createNewGroupButton.style.margin = 'auto'
-          createNewGroupButton.addEventListener('click', _event => {
-            this.createNewGroup(book)
+          createNewGroupButton.addEventListener('click', (_event) => {
+            createNewGroup(book, this.options.defaultNewGroupName)
```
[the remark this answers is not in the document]
SharonStrats: Instead of exporting it I put it back in the class and tested it from there.

package.json
SharonStrats: I'm a bit confused why this shows up in this way here.  it still shows in my local copy. I have changed this however in the other pull request for windows where I change it to \" instead of single quotes.

src/widgets/peoplePicker.js
```diff
@@ -139,30 +139,19 @@ export class PeoplePicker {
     })
   }
 
-  createNewGroup (book) {
+  createNewGroup (book, defaultNewGroupName) {
     const { groupIndex, groupContainer } = indexes(book)
     const group = rdf.sym(
       `${groupContainer.uri}${uuid.v4().slice(0, 8)}.ttl#this`
     )
-    const name = this.options.defaultNewGroupName || 'Untitled Group'
```
SharonStrats: I could put this back, I have left it for your review because I think it's a change we would want to make later anyway.

test/unit/widgets/dragAndDrop.test.ts
```diff
@@ -1,7 +1,11 @@
 import * as RdfLib from 'rdflib'
 import { JSDOM } from 'jsdom'
 
-import { makeDropTarget, makeDraggable, uploadFiles } from '../../../src/widgets/dragAndDrop'
```
SharonStrats: we could remove this file, although it is just a test file that I will change soon as well.

package.json
megoth: I suspect Prettier did a run through the file, and changed format.

SharonStrats: Awww... sorry I didn't think of that.

src/widgets/peoplePicker.js
megoth: I agree

test/unit/widgets/peoplePicker.test.ts
```diff
@@ -5,28 +5,118 @@ import {
...
+  it('should call kb.any and spy load when callback is successful ', () => {
+    const spyOnNowOrWhenFetched = jest.spyOn(fetcher, 'nowOrWhenFetched')
+
+    const spyAny = jest
+      .spyOn(kb, 'any')
+      .mockReturnValueOnce('book')
+      .mockReturnValueOnce('book')
+    const spyLoad = jest.spyOn(fetcher, 'load').mockResolvedValue('book')
+    new PeoplePicker().findAddressBook('typeIndex')
+    const callback: any = spyOnNowOrWhenFetched.mock.calls[0][1]
```
megoth: If you want to callback to follow the signature of findAddressBook, you can consider typing the method and export it so that you could reuse it here (instead of `any`).

Don't need to do it now if you don't think it gives value, just a tip.

test/unit/widgets/peoplePicker.test.ts
```diff
@@ -211,10 +267,40 @@ describe('GroupPicker.loadGroups', () => {
   it('runs', () => {
     const container = RdfLib.sym('')
     const book = RdfLib.sym('')
-    const handler = () => {}
+    const handler = () => { }
     const groupPicker = new GroupPicker(container, book, handler)
     expect(groupPicker.loadGroups()).toBeTruthy()
   })
+  it('should....', () => {
```
megoth: This is not very helpful - please change into something more descriptive.

test/unit/widgets/peoplePicker.test.ts
```diff
@@ -258,10 +348,12 @@ describe('GroupBuilder.render', () => {
   it('exists', () => {
     expect(new GroupBuilder().render).toBeInstanceOf(Function)
   })
+
   it('runs', () => {
+    jest.clearAllMocks()
```
megoth: In general, might want to move this into an `afterEach` block, to make sure it runs after all tests. In this case I don't think it's necessary though, but might want to do it when we add more tests that need to clear all mocks.

Vinnl: I'd set up Jest to do that everywhere - not sure why it's not on by default (since you want unit tests to be independent).

https://jestjs.io/docs/en/configuration#clearmocks-boolean

PR #218, opened 2020-02-24
Show what happens when unit API breaks

michielbdejong:
@megoth see https://github.com/solid/solid-ui/pull/214#discussion_r383173729 -
you said "Do not need tests that check that an imported function exists - TypeScript does this for us."
But this is an export and not an import, right? Or am I looking in the wrong place?

### DIFF
--- a/src/widgets/peoplePicker.js
+++ b/src/widgets/peoplePicker.js
@@ -105,7 +105,7 @@ export class PeoplePicker {
     return this
   }
 
-  findAddressBook (typeIndex) {
+  findAddressBookOops (typeIndex) {
     return new Promise((resolve, reject) => {
       kb.fetcher.nowOrWhenFetched(typeIndex, (ok, err) => {
         if (!ok) {


michielbdejong:
The unit test breaks, but TypeScript doesn't know which methods should exist and which shouldn't.

Of course, we could also describe the class with an interface, that would achieve the same result. But in this case it's just [`export class`](https://github.com/solid/solid-ui/blob/master/src/widgets/peoplePicker.js#L23) so we need to test that, right?

megoth:
This might just be a personal opinion, but if you have tests that covers the use of funtionality, then you have tested that it exists as well. I don't see the value in having a test that checks for the existance of the functionality itself.

Now, how that functionality is exposed publicly for the end user, that is different, and might be something that you want to have an exist test for.

In other words, doing `export foo () {}` and then testing that using something like 

```
import { foo } from <module>

it ('exists', () => {
  expect(foo).toBeInstanceOf(Function)
})
```

Does not make sense if you also have

```
it ('gets an object when calling foo', () => {
  expect(foo()).toBe(<whatever>)
})
```

But checking that foo is exposed as intended might make sense, e.g. in `src/index.ts` you have

```
import { foo } from <module>

export { foo as myMagicFunction }
```

It might makes sense to have the test

```
import * as solidUI from <index>
import { foo } from <module>

it('should expose foo as myMagicFunction', () => {
  expect(solidUI.myMagicFuntion).toBe(foo)
})
```

Makes sense?

michielbdejong:
As agreed on gitter, "if it's already there, let's keep it"

megoth:
One final point: if the time comes that these tests costs us anything in maintanence, I'm going to argue for their removal. But we'll take that discussion if we ever get there.

PR #229, opened 2020-03-01
Migrated tabs.js to TypeScript

megoth:
Can probably be merged directly to `master`, but want to give @SharonStrats the ability to easily see how a migration can be done.


michielbdejong:
Why is this PR still open?

megoth:
We left it open for convenience of @SharonStrats. I don't think it's important for her anymore, so I'll close it now.

Issue #216, opened 2020-02-24
Import from Turtle in Notepad example


michielbdejong:
As discussed with @timbl, https://github.com/solid/solid-ui/blob/master/examples/notepad/index.html#L22-L38 should be an import from a Turtle string, not multiple calls to `store.add`.

jeff-zucker:
See PR #230 and please let me know if there are problems.

michielbdejong:
Great, thank you!

PR #230, opened 2020-03-01
fixes issue #216 (provides pad.ttl for notepad example)

jeff-zucker:


michielbdejong:
Awesome, thank you!!
I tried it out locally and it works a charm. :)

PR #232, opened 2020-03-02
[WIP] Migrating buttons module to ESModules

megoth:
For some reasons tests start breaking in add-agent-buttons


src/widgets/buttons.ts
```diff
@@ -1,4 +1,9 @@
-import { IndexedFormula, NamedNode } from 'rdflib'
+import { IndexedFormula, NamedNode, st, sym, uri, Util } from 'rdflib'
+import icons from '../iconBase'
```
michielbdejong: actually you could even do:
```js
import { iconBase, originalIconBase } from '../iconBase'
```
and search/replace icon.iconBase -> iconBase

PR #233, opened 2020-03-03
[WiP] Converted pad to typescript

SharonStrats:
Completed the typescript and would like to see if I am on the right track.  This one had a few gotchas...

Also, just started adding typedoc...


src/pad.ts
```diff
@@ -281,68 +279,78 @@ UI.pad.notepadToHTML = function (pad, kb) {
...
+  /* @@ TODO want to look into this, it seems upstream should be a boolean and default to false ?
+  *
+  */
+  const complain = function (message: string, upstream: boolean = false) {
     console.log(message)
     if (options.statusArea) {
-      ;(upstream ? upstreamStatus : downstreamStatus).appendChild(
-        UI.widgets.errorMessageBlock(dom, message, 'pink')
-      )
+      if (upstream && upstreamStatus) {
```
SharonStrats: wanted to get this checked as there was an error the old way because it said upstreamStatus and downstreamStatus could both be null.

src/pad.ts
```diff
@@ -676,12 +687,14 @@ UI.pad.notepad = function (dom, padDoc, subject, me, options) {
     return part
   }
 
-  var newChunk = function (ele, before) {
+  /* @@ TODO we need to look at indent, it can be a Number or an Object this doesn't seem correct.
```
SharonStrats: Any advice on indent or should I just leave as is?  I had to change a bit here too because indent could either be a number or an object.

src/pad.ts
```diff
@@ -44,40 +38,40 @@ UI.pad.lightColorHash = function (author) {
 //
 //  This is more general tham the pad.
 //
-UI.pad.renderPartipants = function (dom, table, padDoc, subject, me, options) {
+export function renderPartipants (dom, table, padDoc, subject, me, options) {
```
SharonStrats: padDoc and me are unused.  I didn't want to delete without asking

src/pad.ts
```diff
@@ -28,8 +22,8 @@ const utils = require('./utils')
  * @param {NamedNode} author - The author of text being displayed
  * @returns {String} The CSS color generated, constrained to be light for a background color
  */
-UI.pad.lightColorHash = function (author) {
```
SharonStrats: I still need to add the types to the parameters.  Just realized as there were so many other things to change. I'll have you look at my comments first and after our meeting I'll do the types in the parameters.

src/pad.ts
```diff
@@ -2,24 +2,18 @@
  *   Notepad Widget
  */
 
-/** @module UI.pad
+/** @module pad
  */
+import store from './store'
+import ns from './ns'
```
SharonStrats: widgets, store, ns are missing type declarations.. just a note

megoth: Mm, will be added as we clean them up.

package.json
```diff
@@ -102,4 +102,4 @@
       "eslint"
     ]
   }
-}
\ No newline at end of file
+}
```
megoth: Strange that we get commits such as these =/ Should not happen with proper linting.

src/index.ts
```diff
@@ -76,7 +76,7 @@ import versionInfo from './versionInfo'
 const dom = window ? window.document : null // Idea that UI.dom can be adapted in non-browser environments
 
 if (typeof window !== 'undefined') {
-  ;(<any>window).UI = {
+  ; (<any>window).UI = {
```
megoth: Is this because of ESLint + Standard? Haven't noticed this before, so little strange to see these small variations...

src/pad.ts
michielbdejong: The code assumes that the result of `tr.appendChild(dom.createElement('td'))` is not null, and I think that's a safe assumption, so you can just cast `(upstream ? upstreamStatus : downstreamStatus)` from HTMLElement|null to HTMLElement.

src/pad.ts
```diff
@@ -281,68 +279,78 @@ UI.pad.notepadToHTML = function (pad, kb) {
...
     console.log(message)
     if (options.statusArea) {
-      ;(upstream ? upstreamStatus : downstreamStatus).appendChild(
-        UI.widgets.errorMessageBlock(dom, message, 'pink')
-      )
+      if (upstream && upstreamStatus) {
+        upstreamStatus.appendChild(
+          errorMessageBlock(dom, message, 'pink')
+        )
+      } else if (upstream && downstreamStatus) {
```
michielbdejong: I think you meant `!upstream` :)

src/pad.ts
michielbdejong: Definitely something you should fix as part of the cleanup. In which cases is it a number and in which cases is it an object? When it's an object, then what does that object contain? Can you convert it to a number?

src/pad.ts
michielbdejong: Removing them would mess up the positional arguments of this function.
But you can rename them to `unused1` and `unused2` and mention in the typedoc comment that these are unused positional arguments, left in place for backwards compatibility.

src/pad.ts
SharonStrats: The original code initialized these variables to null.  That's what made me have to do HTMLElement | null.  If I take out the initialization to null this does work though.  Do you think I should go this route (take out the initialization and leave in the upstream ? code)?

michielbdejong: No, you can leave the initialization to null and leave the variable type as `HTMLElement | null` but then cast to `HTMLElement` inside the if statement from line 319, because inside that if you know that lines 303-312 will have been executed earlier.

src/pad.ts
SharonStrats: I see what you are saying. I will see if I can determine this as I write my examples.

src/pad.ts
SharonStrats: will do.

src/pad.ts
SharonStrats: cool.  will do :).

src/pad.ts
SharonStrats: oops will change this.

src/pad.ts
SharonStrats: Added a comment for now in the code.

examples/notepad/index.html
```diff
@@ -1,33 +1,157 @@
 <!DOCTYPE HTML>
 <html lang="en">
-  <head>
```
SharonStrats: Can you have a quick look at what I did here?  I'm not completely done.  I wish it were shown a bit more like a "real" example...  just want to get your thoughts?

michielbdejong: Why did you unindent this? the `<head>` tag comes under the `<html>` tag, right?

SharonStrats: hmm, don't think it was intentional.  I will take a look

src/pad-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+export function xmlEncode (str) {
+  return str.replace('&', '&amp;').replace('<', '&lt;').replace('>', '&gt;')
+}
+
+/**
+ * Convert a notepad to HTML
+ *   @param { } pad - the notepad
+ *   @param {store} pad - the data store
+ */
+export function notepadToHTML (pad: any, kb: store) {
```
SharonStrats: I am not 100% sure how to add this to index.ts?

test/unit/pad-utils.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import { getChunks, notepadToHTML, xmlEncode } from '../../src/pad-utils'
+
+describe('getChunks', () => {
+  it('exists', () => {
+    expect(getChunks).toBeInstanceOf(Function)
+  })
+})
+
+describe('notepadToHTML', () => {
+  it('exists', () => {
```
megoth: Please lets stop writing these tests - TypeScript does this for us. Write tests that make sure that the functions are exported correctly to the API of Solid UI, e.g. check out what I did for [tabWidget in the tabs module](https://github.com/solid/solid-ui/pull/247/files#diff-2d70df5dd69be0ab433b268762ec8b97R27).

test/unit/pad.test.ts
```diff
@@ -1,192 +1,77 @@
...
+    expect((pad as any).notepad(dom, padDoc, subject, me, options))
+      .toMatchSnapshot()
+  })
+
+  it.skip('should log error that you need to be logged in for pad to be edited', () => {
+    const padDoc = null
+    const subject = null
+    const me = null
+    const options = {}
+      ; (window as any).console = { log: jest.fn() }
```
megoth: I think [spyOn](https://jestjs.io/docs/en/jest-object.html#jestspyonobject-methodname) is a better to use here, e.g. `jest.spyOn(window.console, 'log')`

test/unit/participation.test.ts
```diff
@@ -0,0 +1,97 @@
+import widgets from '../../src/widgets'
+import { JSDOM } from 'jsdom'
+import * as participation from '../../src/participation'
```
megoth: Why do you pull in the participation module like this? Why not import the methods directly, e.g. `import { renderPartipants, participationObject, recordParticipation, manageParticipation } from '../../src/participation'`?

megoth: Also, if you do need to do this (which you might want to do because you want to test against the public API, like I do in [tabs test](https://github.com/solid/solid-ui/pull/247/files#diff-2d70df5dd69be0ab433b268762ec8b97R27)), you should _not_ need to do `(participation as any)` - we've created those APIs, not reason to throw it to any type).

michielbdejong:
Feedback on the examples:
* the result of lightColorHash is a string, but it's then converted into a Notepad for displaying. That's confusing. You can just display the string by itself using `document.createTextNode`?
* renderParticipants example is broken (it shows a notepad containing the string '[object HTMLTableElement]')
* the others haven't been done yet, apparently? would be good to add a "FIXME: example missing" and create a follow-up ticket for each one. You can add the 'good first issue' label to attract community contributions

test/unit/participation.test.ts
SharonStrats: hmm.. I will look into this.  I didn't understand the difference and just saw that in another place it was done this way. I will change this and try to understand the difference. Meaning the "why" to use it.

test/unit/pad-utils.test.ts
SharonStrats: I see what you've done here...  on this one in particular getChunks it is not exposed on the public api.  I guess I should just check that it runs and the functionality within?

src/pad-utils.ts
SharonStrats: Can we go over this and also the participation after our standup?  I just have a couple of questions.

examples/notepad/index.html
SharonStrats: on my screen it does not show as indented.  I will resolve this and see if it appears again.

src/pad.ts
```diff
@@ -9,6 +9,8 @@ import ns from './ns'
 import { Namespace, NamedNode, st } from 'rdflib'
 import { newThing, errorMessageBlock } from './widgets'
 import { beep } from './utils'
+export { notepadToHTML } from './pad-utils'
```
megoth: I think this method should be moved into this file. (I don't think any function that are exposed to the public API should be in `-utils`-files.)

src/pad.ts
```diff
@@ -9,6 +9,8 @@ import ns from './ns'
...
+export { renderPartipants, participationObject, manageParticipation, recordParticipation } from './participation'
```
megoth: This is important wrt backward-compatibility - I agree that we can have this until we do a breaking change version of solid-ui. Might want to add a comment wrt to this.

michielbdejong:
@SharonStrats I had another look at the examples, maybe you have some unpushed changes, but they all seem to do the iframe thing that the notepadToHTML uses, rather than showing how to use the function in question, for instance:
Screenshot 2020-03-11 11 36 13


So I had a go at fixing the example for the main `UI.pad.notepad` function, see https://github.com/solid/solid-ui/pull/251 - do you think you could make the other examples more like that?

test/unit/pad-utils.test.ts
megoth: Yeah, if it's an internal method, you want to check that it does what you intend, no need to check the public API.

michielbdejong:
@SharonStrats as discussed, let's try to get #273 and #274 done this sprint, and then leave the code changes (typescript + typedoc) and the userguide page for next sprint.

SharonStrats:
Just wanted to leave a note here to say: This PR has all the typescript and typedoc work for PAD, we are on onhold with this until we complete better pad unit tests.  Then this PR can be completed.

SharonStrats:
Decided to redo since it was 173 commits behind.

PR #235, opened 2020-03-04
Changed callback to be named

SharonStrats:
Had to change callback to be a named function.


SharonStrats:
I was thinking the same thing. At the time I received this error it was unexpected and I needed to fix this file to get my draganddrop pushed.  I am going to revisit the tests and will change it.

michielbdejong:
That's weird because the tests were passing in master without this change, right?

Why was this change necessary, do we know?

SharonStrats:
This happened at one point when there was a problem with buttons.  I can't say that it was necessary once that was fixed, I can test this.  I do think it's probably better to have not named them callback though since it does work with a descriptive name. Should I test to be sure?

michielbdejong:
nah, just leave it now.

PR #237, opened 2020-03-04
Changing how mocking of rdflib works + fixing most tests

megoth:
This allows for reusing most of rdflib.js as is, with the exception of Fetcher and UpdateManager (the latter is dependent on the former).

I've fixed most tests that broke, with exception of peoplePicker.test.ts, pad.test.ts and forms.test.ts, as these are being currently worked on by others. I've simply commented this out, and they should prioritize their own work when merge conflict happen, and then fix the broken tests that may have come.

Note: We might want to change UpdateManager in rdflib.js to optionally take in Fetcher so that we don't need to mock it.


test/unit/chat/dateFolder.test.ts
```diff
@@ -1,87 +1,54 @@
...
-
 import DateFolder from '../../../src/chat/dateFolder'
-jest.mock('rdflib')
+
 jest.mock('solid-auth-client')
 
 describe('DateFolder', () => {
   it('exists', () => {
-    expect(new DateFolder({ dir: () => {} })).toBeInstanceOf(DateFolder)
+    expect(new DateFolder(RdfLib.sym('https://domain.tld/dir/'))).toBeInstanceOf(DateFolder)
```
michielbdejong: remove `RdfLib.` like in other files

test/unit/helpers/clearStore.ts
```diff
@@ -0,0 +1,5 @@
+import store from '../../../src/store'
```
michielbdejong: this uses the store from '../../../src/store' but helpers/createLiveStore.ts does not, how does that work?

test/unit/widgets/__snapshots__/peoplePicker.test.ts.snap
```diff
@@ -1,143 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
michielbdejong: Did you delete this on purpose?

tsconfig.json
```diff
@@ -68,6 +68,7 @@
   },
   "include": [
     "src/**/*",
-    "test/**/*"
+    "test/**/*",
+    "__mocks__/*"
```
michielbdejong: :+1:

test/unit/chat/dateFolder.test.ts
megoth: Ok, that will require a bit more changes, as functions in dateFolder.js is dependent on the rdflib being exposed on the window-object. But it's good to write this away, so will do that.

test/unit/helpers/clearStore.ts
megoth: `createLiveStore` creates a new instance and can be used to inject a mocked version of the store.

The store-reference is global and should be "cleaned" (e.g. by using clearStore-helper) after each test that adds triples to it.

I prefer the former as it cannot lead to "bleeding tests" (i.e. tests that forgot to clean up after themselves), but we cannot use it all the time, as some functionality depends on the global store.

Thumb finger rule: If you can inject the store that should be worked on, prefer adding an instance created with `createLiveStore`.

test/unit/widgets/__snapshots__/peoplePicker.test.ts.snap
megoth: Yes, because it's dependent on tests in peoplePicker.test.ts. I've commented out all the tests in peoplePicker.test.ts for now, to not interfere with the work of @SharonStrats. She'll simply add the snapshot anew when she fixes peoplePicker.test.ts after rebasing on master.

PR #247, opened 2020-03-09
Tabs cleanup

megoth:
This PR is only for sharing progress so far (and being able to show to certain solutions I've chosen).

Do not review yet.


.babelrc
```diff
@@ -5,6 +5,7 @@
   ],
   "plugins": [
     "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-proposal-optional-chaining",
```
megoth: Could push this into a separate commit if wanted - I introduced the use of [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) when working on tabs.ts.

michielbdejong:
http://localhost:5000/test/browser/tabs/ works
http://localhost:5000/examples/tabs/ works
Branch coverage in `src/tabs.ts` is 86%, w00t!

examples/tabs/index.html
```diff
@@ -1,582 +1,402 @@
...
+<article id="tabWidget-with-collection">
+    <h2>
+        <a href="#tabWidget-with-collection">
+            <code>tabWidget</code>: Populating content using a RDF collection
+        </a>
+    </h2>
+    <p>
+        In this first example we'll see how you can generate a tabs widget using triples that's formatted as a
+        collection
+        in RDF. See <a href="http://127.0.0.1:8080/Documentation/api/modules/_tabs_.html#tabwidget">API
```
michielbdejong: Link to https://solid.github.io/solid-ui/Documentation/api/modules/_tabs_.html#tabwidget instead

examples/tabs/index.html
```diff
@@ -1,582 +1,402 @@
...
+<h1>The Tabs module</h1>
+
+<p>
+    On this page you'll find some examples from the Tabs module.
+</p>
+
+<article id="tabWidget-with-collection">
+    <h2>
+        <a href="#tabWidget-with-collection">
+            <code>tabWidget</code>: Populating content using a RDF collection
```
michielbdejong: "using a*n* RDF collection"

examples/tabs/index.html
```diff
@@ -1,582 +1,402 @@
...
+        </a>
+    </h2>
+    <p>
+        In this first example we'll see how you can generate a tabs widget using triples that's formatted as a
```
michielbdejong: The triples are what is formatted, right? Then you should say "triples that are" instead of "triples that's"

examples/tabs/index.html
```diff
@@ -1,582 +1,402 @@
...
+          predicate,
+          renderMain: (bodyMain, subject) => {
+            bodyMain.innerText = `Content of ${subject.uri}`
+          }
+        })
+        document.querySelector('#demo-tabWidget-with-collection').appendChild(tabs)
+      })
+    </script>
+    <p>
+        The triples belows shows the minimum data you need for to start using the tabWidget in combination with
```
michielbdejong: "you need ~for~ to start using"

examples/tabs/index.html
```diff
@@ -1,582 +1,402 @@
...
+        UI.store.
+    </p>
+    <pre id="viewData-tabWidget-with-collection"></pre>
+    <p>
+        Note that you would probably use UI.rdf.namedNode directly from rdflib.js, but we've used the convenience
+        reference
```
michielbdejong: which convenience reference? Looks like you do use UI.rdf.namedNode

examples/tabs/index.html
```diff
@@ -1,582 +1,402 @@
...
+</article>
+
+<article id="tabWidget-option-backgroundColor">
+    <h2>
+        <a href="#tabWidget-option-backgroundColor">
+            <code>tabWidget</code>: Option <code>backgroundColor</code>
+        </a>
+    </h2>
+    <p>
+        Using option <code>backgroundColor</code> we can set a specific color (in hex value) to our tabs. It will
```
michielbdejong: Mention that this is the unselected background colour, selected will be darker

examples/tabs/index.html
```diff
@@ -1,582 +1,402 @@
...
+        const demo = document.querySelector('#demo-tabWidget-option-renderTab')
+        const item1 = UI.rdf.namedNode(location.origin + '#item3A')
+        const item2 = UI.rdf.namedNode(location.origin + '#item3B')
+        demo.appendChild(UI.tabs.tabWidget({
+          items: [item1, item2],
+          renderMain: (bodyMain, subject) => {
+            bodyMain.innerText = `Content of ${subject}`
+          },
+          renderTab: (tabDiv, subject) => {
+            tabDiv.innerText = subject.uri
```
michielbdejong: maybe do something different here from the default, to show the effect, e.g.:
```js
tabDiv.innerText = `Go to ${subject.uri}!`
```

examples/tabs/index.html
```diff
@@ -1,582 +1,402 @@
...
+    <div id="demo-tabWidget-option-renderTab"></div>
+</article>
+
+<article id="tabWidget-option-renderTabSettings">
+    <h2>
+        <a href="#tabWidget-option-renderTabSettings"><code>tabWidget</code>: Option <code>renderTabSettings</code></a>
+    </h2>
+    <p>
+        <code>renderTabSettings</code> functions much like <code>renderMain</code>, expect that it is triggered by
+        holding ALT key and clicking on a tab.
```
michielbdejong: (or 'option' on a Mac)

michielbdejong: and wow, never knew this functionality existed! :)

michielbdejong:
Comments are all minor, you're good to merge with or without them as you see fit.

src/tabs.ts
```diff
@@ -0,0 +1,412 @@
...
+ * @internal
+ */
+function getColors (backgroundColor: string): [string, string] {
+  return isLight(backgroundColor)
+    ? [colorBlend(backgroundColor, '#ffffff', 0.3), '#000000']
+    : [colorBlend(backgroundColor, '#000000', 0.3), '#ffffff']
+}
+
+/**
+ * @internal
```
SharonStrats: just wondering why you used @internal instead of @ignore? I can't seem to find anything on @internal to see why to use one versus the other.

megoth: There's a [section on @internal in the TypeDoc guides](https://typedoc.org/guides/doccomments/#internal), and I thought the distinction might be good to have (they are, after all, internal functions that should not be exposed on the public API), especially since TypeDoc has the option `stripInternal`.

Now, if we want to describe internal APIs as well at some point, we can just remove the option `stripInternal` when generating the docs.

examples/tabs/index.html
megoth: `UI.rdf` _is_ the convenience reference. Maybe convenience reference isn't the best word?

megoth: I'll remove the word convenience reference and be a bit more explicit about what you would do instead.

examples/tabs/index.html
megoth: Heh, yeah, there are some magic in the existing solution.

PR #251, opened 2020-03-11
Fix notepad example

michielbdejong:


michielbdejong:
This uses a public notepad at https://example-user.inrupt.net/public/example-notepad/index.ttl which is susceptible to vandalism and definitely not ideal.

I tried to insert triples into UI.store to avoid it fetching the document, but it seemed to try to fetch the doc anyway, so not sure how to make this example more resistant to vandalism.

Also, when I type some lines and then reload the page, the lines I typed disappear, not sure why that is?

michielbdejong:
Screenshot 2020-03-11 12 27 20

michielbdejong:
Ah looking at the Turtle in https://example-user.inrupt.net/public/example-notepad/index.ttl it looks like the reason it's not saving the edits is that example is incorrect, should probably be:
```js
const result = UI.pad.notepad(document, pad.doc(), subject, me, options)
```
instead of:
```js
const result = UI.pad.notepad(document, pad, subject, me, options)
```

michielbdejong:
@SharonStrats I'll leave it to you to figure out the rest of that, if that's OK!

michielbdejong:
This was merged and is now part of #274.

michielbdejong:
@SharonStrats:
> when you created example-notepad under example-user did you do anything more than use UI.pad.notepad.. ? meaning did you create it on the pod first?

Yeah, I created it on the pod first.

SharonStrats:
just the folder? example-notepad

SharonStrats:
got it.  Yes so create a notepad under public called example-notepad

michielbdejong:
> creating an input field for the user to be able to enter the name of their pad.

Why not just use http://solid.github.io/solid-ui/examples/buttons/#askName for that? It's exactly what that component was meant to do! :)
You can also have a look at src/create.js

Hm, but wait, will your example then require the user to log in? Probably yes, right? Because if not, then creating a Notepad would only be allowed in a public-writable folder, and I don't know what's worse, pointing to a pre-created world-writable Notepad doc, or pointing to a pre-created world-writable container. :)

So you can either stick with the way I did it, or if you have time left in the sprint, you can create the whole log-in-and-create-an-empty-notepad-on-your-own-pod flow, I don't think it's a hard requirement for your Notepad example but definitely a nice-to-have!

SharonStrats:
I will look into this.  I was thinking about the whole logging in thing.... that's a good point.

PR #255, opened 2020-03-13
Re-enable forms tests

michielbdejong:


test/unit/widgets/forms.test.ts
```diff
@@ -1,1435 +1,1455 @@
...
+      field[ns.ui('Form').uri](
+        document,
+        container,
+        already,
+        subject,
+        form,
+        store,
+        callbackFunction
+      )
+    ).toMatchInlineSnapshot(`
```
megoth: Please replace all instances toMatchInlineSnapshot with toMatchSnapshot.

(And also remove the use of Prettier - don't know how that got back in there.)

test/unit/widgets/forms.test.ts
```diff
@@ -1,1435 +1,1455 @@
...
+  it('exists', () => {
+    expect(field[ns.ui('Multiple').uri]).toBeInstanceOf(Object)
+  })
+  it.skip('runs', () => {
+    const container = document.createElement('div')
+    const already = {}
+    const subject = namedNode('http://example.com/#this')
+    const form = namedNode('http://example.com/#form')
+    const store = namedNode('http://example.com/#store')
+    const callbackFunction = () => {}
```
megoth: I see a lot of these in the test file, but seldom actually a test checking when it's run. Please write tests that check when they are called.

test/unit/widgets/forms.test.ts
```diff
@@ -1,1435 +1,1455 @@
...
+  })
+})
+
+describe('fieldParams', () => {
+  it('exists', () => {
+    expect(fieldParams).toBeInstanceOf(Object)
+  })
+})
+
+describe('ColorField params', () => {
```
megoth: Maybe you should group this under describe('fieldParams') block? Since they are all related to `fieldParams`, makes it easier to scan the blocks of tests.

test/unit/widgets/forms.test.ts
```diff
@@ -1,1435 +1,1455 @@
...
+        subject,
+        form,
+        store,
+        callbackFunction
+      )
+    ).toMatchInlineSnapshot(`
+<div>
+        <a
+          href="http://example.com/#bla"
+          style="color: #3B5998; text-
```
megoth: Would be good to have a test when this callback is called. Similar with all the other tests that simply give a `() => {}` in the callback parameter.

You probably want to split the test into two tests, one to check the instance, and another to check that the callback was called when expected. Use `beforeEach` to group test code that is called both places.

test/unit/widgets/forms.test.ts
michielbdejong: yeah, they're moved to a separate file in the fieldParams PR

test/unit/widgets/forms.test.ts
michielbdejong: This PR only re-enables the tests that you deleted when you switched rdflib from a mock to a spy.
The existing tests only test how form fields are created, not how you can interact with them, that's why none of those callbacks actually get called in any of the tests.

I created a follow-up issue https://github.com/solid/solid-ui/issues/263 so that we can merge this and add the interaction tests in additional set of tests.

test/unit/widgets/forms.test.ts
michielbdejong: See https://github.com/solid/solid-ui/pull/255#discussion_r392251757

PR #258, opened 2020-03-13
Clean up errorMessageBlock

michielbdejong:


michielbdejong:
This file just had 18 lines of code so I quickly cleaned it up.

test/unit/widgets/error.test.ts
```diff
@@ -5,11 +5,17 @@ import { errorMessageBlock } from '../../../src/widgets/error'
...
 
-describe('button', () => {
+describe('errorMessageBlock', () => {
   it('exists', () => {
     expect(errorMessageBlock).toBeInstanceOf(Function)
   })
-  it('runs', () => {
-    expect(errorMessageBlock(dom, '', undefined)).toBeTruthy()
+  it('creates an error message div', () => {
+    expect(errorMessageBlock(dom, 'my error message', undefined)).toMatchSnapshot()
```
megoth: Don't need to set the third parameter when it's undefined (that's the default behavior, after all).

test/unit/widgets/error.test.ts
```diff
@@ -5,11 +5,17 @@ import { errorMessageBlock } from '../../../src/widgets/error'
...
+  })
+  it('escapes the HTML', () => {
+    expect(errorMessageBlock(dom, 'my <script> kiddie </script> message', undefined)).toMatchSnapshot()
```
megoth: Might want to check this more explicitly (e.g. using toMatch and making sure that < and > is not present in text), since a snapshot doesn't really test this in a good way I think.

PR #260, opened 2020-03-13
Fix #259

### LINKED ISSUE #259: fieldFunction does not append error block to container
If a good element can't be found for a given field, an error div is returned, but the code to add that div to the container if present is missing.

michielbdejong:


michielbdejong:
This also makes the error function use its own reference to the DOM, so that means the first argument in fieldFunction becomes unused (but I'll leave it there for backwards compat)

src/widgets/forms/fieldFunction.ts
```diff
@@ -25,26 +25,30 @@ export function mostSpecificClassURI (x: Node): string {
   const kb = store
   const ft = kb.findTypeURIs(x)
   const bot = kb.bottomTypeURIs(ft) // most specific
+  console.log(x, ft, bot)
```
megoth: Forgot to remove?

michielbdejong:
OK I'll just mark it as unused in the TypeDoc comments then.

michielbdejong:
> unsure whether renaming it to unused is a good pattern, maybe something we should raise an issue for?

OK I'll just mark it as unused in the TypeDoc comments for now then.

PR #264, opened 2020-03-13
Split out and clean up mostSpecificClassURI, fieldFunction

michielbdejong:
100% branch coverage, move-to-TS, examples, and typedoc for both functions.

Also simplified the other occurrences of () => { clearStore() } to just clearStore.


src/widgets/forms/fieldFunction.ts
```diff
@@ -0,0 +1,60 @@
+import { Node } from 'rdflib'
+import store from '../../store'
+import { debug } from '../../log'
+import { errorMessageBlock } from '../error'
+
+export const field: { [classUri: string]: FieldFunction } = {} // Form field functions by URI of field type.
```
megoth: Maybe you want to add [the entries to forms.field from forms.js](https://github.com/solid/solid-ui/blob/fieldFunction/src/widgets/forms.js#L37-L537) in this file as well? 

You could construct a map on the top, with pointers to the functions, either in this file or in separate files if you deem them so complex that it makes more sense to split them off into separate modules.

michielbdejong: Yes! Will do that in a follow-up.

PR #267, opened 2020-03-13
Enable https://eslint.org/docs/rules/no-console

michielbdejong:


src/debug.ts
```diff
@@ -0,0 +1,18 @@
+export const debug = {
```
michielbdejong: This seems to work, right?

michielbdejong:
@megoth as discussed, let's wait for you to discuss this with @timbl before approving.

michielbdejong:
@megoth any progress on this? I would like to get this merged before the end of the week, so it doesn't drag on too much.

src/acl/acl.ts
```diff
@@ -373,7 +374,7 @@ export function fixIndividualCardACL (person: $rdf.NamedNode, log: Function, cal
  * This function is used by [[fixIndividualCardACL]]
  */
 export function fixIndividualACL (item: $rdf.NamedNode, subjects: Array<$rdf.NamedNode>, log: Function, callbackFunction: Function): void {
-  log = log || console.log
+  log = log || debug.log
```
megoth: Think this could be set as default param for the log parameter, but of course that begs to ask what we should set as default value for callbackFunction. Nevertheless something we want to clean up later, to not make this PR too extensive.

src/debug.ts
megoth: You could also just export each funtion separately, e.g. `export function log (...args: any[])` instead of packing it up in an object.

src/debug.ts
```diff
@@ -0,0 +1,18 @@
...
+  log (...args: any[]) {
+    // eslint-disable-next-line no-console
+    console.log(...args)
```
megoth: Had forgotten that you can spread arguments to a function call that way. Nifty.

src/widgets/index.js
```diff
@@ -28,6 +30,6 @@ var widgets = Object.assign(
 )
 
 module.exports = widgets
-console.log('@@@ widgets: ', widgets)
+debug.log('@@@ widgets: ', widgets)
```
megoth: We need to check with Tim how valuable this line actually is currently. It's really annoying in tests.

test/unit/helpers/clearStore.ts
```diff
@@ -1,9 +1,9 @@
 import store from '../../../src/store'
```
megoth: Lets revert the changes to this file? Think we have a much better solution in master branch.

src/widgets/index.js
michielbdejong: Agree! :)

src/debug.ts
michielbdejong: Done in c5f47f0

PR #269, opened 2020-03-16
Log cleanup

megoth:
This PR includes migration to TS, unit tests, documentation, and examples.


michielbdejong:
Great! Will review after the standup.

src/log.ts
```diff
@@ -0,0 +1,189 @@
...
+ * - Warning: 2
+ * - Message: 4
+ * - Success: 8
+ * - Info: 16
+ * - Debug: 32
+ *
+ * You can also choose to set all by passing the number *64*.
+ */
+export function setLevel (level: number): void {
+  _level = TALL
```
michielbdejong: What does this line do?

src/log.ts
```diff
@@ -0,0 +1,189 @@
...
+ * sum of the levels you want to show:
+ *
+ * - Error: 1
+ * - Warning: 2
+ * - Message: 4
+ * - Success: 8
+ * - Info: 16
+ * - Debug: 32
+ *
+ * You can also choose to set all by passing the number *64*.
```
michielbdejong: Wouldn't 63 make more sense instead of 64?

examples/log/index.html
```diff
@@ -0,0 +1,295 @@
...
+    <ul>
+        <li>Error: 1</li>
+        <li>Warning: 2</li>
+        <li>Message: 4</li>
+        <li>Success: 8</li>
+        <li>Info: 16</li>
+        <li>Debug: 32</li>
+    </ul>
+    <p>
+        Setting level to 64 will show all messages.
```
michielbdejong: 63, not 64. You could also clarify "It's a bitmask, so setting level to 63 (i.e. 1+2+4+8+16+32) will show all messages."

src/log.ts
megoth: Oh, yes, my bad, will fix

src/log.ts
megoth: Just making sure that the next line are shown, could be set to TDEBUG instead.

examples/log/index.html
megoth: Yeah, was wondering whether or not to mention the bitmask thing, as some might be confused by it. But I'll phrase it in such a way that is useful for those that understand and not necessary to understand for those who don't.

PR #275, opened 2020-03-19
Basic field cleanup

michielbdejong:


src/widgets/forms/basic.ts
```diff
@@ -0,0 +1,235 @@
...
+import { fieldParams } from './fieldParams'
+
+/**
+ * Create an anchor element with a label as the anchor text.
+ *
+ * @param dom The DOM
+ * @param property href for the anchor element
+ * @param fieldInQuestion field to produce a label for
+ *
+ * @ignore exporting this only for unit tests
```
megoth: Any reason why you don't use @internal?

src/widgets/forms/basic.ts
```diff
@@ -0,0 +1,235 @@
...
+ * @param predicate Predicate about which we want to find an editable RDF document
+ * @param def default RDF document to return if none found
+ *
+ * @ignore exporting this only for unit tests
+ */
+export function fieldStore (subject: Node, predicate: Node, def: Node | undefined): Node | undefined {
+  const sts = store.statementsMatching(subject, predicate)
+  if (sts.length === 0) return def // can used default as no data yet
+  if (
+    sts.length > 0 &&
```
megoth: That's kinda implied by line 47, ain't it?

michielbdejong: yeah, I didn't do any refactor in this branch. Now that we have full branch coverage, it's definitely something I want to do in a next PR!

src/widgets/forms/basic.ts
michielbdejong: ah right I keep forgetting which one is which :)

PR #280, opened 2020-03-22
Fixing jest config + Ignoring debug methods for tests

megoth:
Realized that jest config must have file-extension .js in order to let jest recognize it properly. Cleaned it up a bit and made it so that calls on debug methods are ignored in tests.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  verbose: true,
+  setupFilesAfterEnv: [
+    './test/setup.ts'
+  ]
+}
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,6 +1,3 @@
 module.exports = {
-  verbose: true,
-  setupFilesAfterEnv: [
-    './test/setup.ts'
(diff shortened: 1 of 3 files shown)


megoth:
Ok, I see your point about not understanding why the debugging module won't work as intended. I'll change the behavior to make it explicit, and then update this PR.

michielbdejong:
Mainly that each file under test/unit/ contains *something* at the top that points to the fact that debug output is disabled.

megoth:
I realize that for this to have the intended functionality, it needs to be run before all tests (as part of `setupFilesAfterEnv` in jest.config.js), since some calls on debug.log are done on the top-level of modules (e.g. in [src/widgets/index.js](https://github.com/solid/solid-ui/blob/master/src/widgets/index.js#L33)).

This was more of a convenience, since I'm buggered by the console.logs every time I run tests. But adding this might cause us to not clean up the use of console.logs anyway, so I'll remove it for now, simply leaving the filename change, which I'll just merge immediately.

PR #283, opened 2020-03-23
Acl cleanup

megoth:
Some work that I started doing for cleaning up the ACL module. It is not complete, but I think it adds value to the code base.

I hope the added matchers are reasonable additions to the test suite.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,6 @@
 module.exports = {
-  verbose: true
+  verbose: true,
+  setupFilesAfterEnv: [
+    './test/setup.ts'
+  ]
 }
--- a/test/setup.ts
+++ b/test/setup.ts
@@ -1,13 +1,16 @@
 import { toContainGraph } from './custom-matchers/toContainGraph'
 import { toEqualGraph } from './custom-matchers/toEqualGraph'
+import { error, log, trace, warn } from '../src/debug'
 
 // We don't want to output debug messages to console as part of the tests
 jest.mock('../src/debug')
-const debug = require('../src/debug')
-debug.log.mockImplementation(() => null)
-debug.warn.mockImplementation(() => null)
-debug.error.mockImplementation(() => null)
-debug.trace.mockImplementation(() => null)
+
+export function silenceDebugMessages () {
+  ;(log as any).mockImplementation(() => null)
+  ;(warn as any).mockImplementation(() => null)
+  ;(error as any).mockImplementation(() => null)
+  ;(trace as any).mockImplementation(() => null)
+}
 
 // adding custom matchers
 expect.extend({
--- a/test/setup.ts
+++ b/test/setup.ts
@@ -1,11 +1,16 @@
-import { log, warn, error, trace } from '../src/debug'
+import { toContainGraph } from './custom-matchers/toContainGraph'
+import { toEqualGraph } from './custom-matchers/toEqualGraph'
 
 // We don't want to output debug messages to console as part of the tests
 jest.mock('../src/debug')
+const debug = require('../src/debug')
+debug.log.mockImplementation(() => null)
+debug.warn.mockImplementation(() => null)
+debug.error.mockImplementation(() => null)
+debug.trace.mockImplementation(() => null)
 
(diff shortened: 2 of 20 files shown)


jest.config.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
-  verbose: true
+  verbose: true,
+  setupFilesAfterEnv: [
+    './test/setup.ts'
+  ]
```
michielbdejong: Seems your change from https://github.com/solid/solid-ui/pull/280 accidentally made it in here? Please remove.

test/setup.ts
```diff
@@ -0,0 +1,8 @@
+import { toContainGraph } from './custom-matchers/toContainGraph'
+import { toEqualGraph } from './custom-matchers/toEqualGraph'
+
+// adding custom matchers
+expect.extend({
+  toContainGraph,
+  toEqualGraph
+})
```
michielbdejong: OK to add this file, but please call it from individual test files, not from jest config.

test/unit/acl/acl.test.ts
```diff
@@ -22,230 +22,374 @@ import {
...
+// })
+//
+// describe('ACLToString', () => {
+//   it('exists', () => {
+//     expect(ACLToString).toBeInstanceOf(Function)
+//   })
+//   it('runs', () => {
+//     expect(ACLToString({} as AgentMapMap)).toEqual('{}')
+//   })
+// })
```
michielbdejong: Why did you comment this out? Committed by accident?

test/unit/acl/acl.test.ts
```diff
@@ -22,230 +22,374 @@ import {
...
+// describe('fixIndividualCardACL', () => {
+//   it('exists', () => {
+//     expect(fixIndividualCardACL).toBeInstanceOf(Function)
+//   })
+//   it('runs', () => {
+//     expect(fixIndividualCardACL(sym('https://test.test#'), () => {
+//     }, () => {
+//     })).toEqual(undefined)
+//   })
+// })
```
michielbdejong: Please uncomment and where necessary use .skip

test/unit/acl/acl.test.ts
```diff
@@ -22,230 +22,374 @@ import {
...
+//       createLiveStore(),
+//       sym('https://test.test#'),
+//       {} as AgentMapMap,
+//       sym('https://test.test#'),
+//       () => {
+//       }
+//     )).toEqual(undefined)
+//   })
+// })
+
```
michielbdejong: Please uncomment and where necessary use .skip

michielbdejong:
Not sure why Trellis is yellow, it shows green when you follow the link.

test/setup.ts
megoth: I'm uncertain of how much value it is to make this opt-in? Yes, there's the problem of discoverability, perhaps I should add a section to developer docs describing these custom matchers instead?

jest.config.js
megoth: See https://github.com/solid/solid-ui/pull/283#discussion_r397188875

test/unit/acl/acl.test.ts
megoth: Sorry about that, forgot to readd tests that I didn't have time to complete. I've readded the tests in my latest commit.

test/setup.ts
michielbdejong: See https://github.com/solid/solid-ui/pull/285

megoth: Hence my point on discoverability. I don't think we should make custom matchers opt-in though, they should always be available to all tests.

michielbdejong: We're commenting on a commit that doesn't exist anymore? My original comment was about silencing the debug messages, I didn't intend to say anything about custom matchers.

Sounds like yeah, these matchers should always be available to all tests.

megoth: Then you commented the wrong place ;) Or maybe we're seeing different things:

Screen Shot 2020-03-25 at 17 54 39

Ok, but then this PR should be ok, and I'll merge it as is.

PR #289, opened 2020-03-26
Sharon's work on buttons

michielbdejong:


src/widgets/buttons.ts
```diff
@@ -477,24 +488,62 @@ export function deleteButtonWithCheck (
   return deleteButtonElt
 }
 
+function getButtonStyle (options: ButtonWidgetOptions) {
+  const color = '#7c4dff'
+
+  const backgroundColor = options.filled ? color : '#ffffff'
```
SharonStrats: this is just the beginning I'm going to make this better by adding a case statement and also need to look deeper into the style so that it is the same as the SDK

src/widgets/buttons.ts
```diff
@@ -26,6 +27,16 @@ export type StatusAreaContext = {
   div?: HTMLElement
   dom?: HTMLDocument
 }
+enum ButtonType {
```
SharonStrats: I realized we already had two other cases in the examples both the cancel and continue buttons use the button widget so I think we should consider these styles as well.

src/widgets/buttons.ts
SharonStrats: Actually maybe even separate functions for each buttonType....  but I am not sure if that's going too far.

megoth:
Should I review this even if it's prefixed with [WiP]?

SharonStrats:
Arne, not review but maybe just have a glance at my comments and see what you think.

michielbdejong:
I'll pauze my work on #291 until @SharonStrats is done with these additions.

src/widgets/buttons.ts
```diff
@@ -27,6 +27,16 @@ export type StatusAreaContext = {
   div?: HTMLElement
   dom?: HTMLDocument
 }
+export enum ButtonType {
+  Primary = '#7c4dff',
+  Secondary = '#01C9EA',
+  Cancel = '',
```
SharonStrats: You can see below there is a cancelButton and a continueButton that literally just call button with Cancel and Continue should we look at setting these styles?

src/widgets/buttons.ts
```diff
@@ -27,6 +27,16 @@ export type StatusAreaContext = {
...
+  Continue = ''
+}
+export type ButtonWidgetOptions = {
+  buttonType?: ButtonType,
```
SharonStrats: I would've like to make this mandatory, but so that I didn't break code I made them optional and then defaulted the style to Primary, filled.  Any objections?

src/widgets/buttons.ts
```diff
@@ -478,6 +488,42 @@ export function deleteButtonWithCheck (
...
 }
 
+function getButtonStyle (options: ButtonWidgetOptions = { buttonType: ButtonType.Primary, filled: true }) {
+  // We need to accomadate for legacy code, which is why we have to allow buttonType and filled to be optional
+  let backgroundColor: (ButtonType | string | undefined) = options.buttonType
+  let fontColor: (ButtonType | string | undefined) = '#ffffff'
+  let borderColor: (ButtonType | string | undefined) = options.buttonType
+  let hoverBackgroundColor: (ButtonType | string | undefined) = `lighten(${options.buttonType}, 5%)`
+  let hoverFontColor: (ButtonType | string | undefined) = fontColor
+  if (!options.filled) {
```
SharonStrats: I went back and forth with this.  At one point I had let setting defaults and then have if (filled), else but decided it was actually adding much more code than necessary and maybe the better approach would be to just default to what the values would be if the option filled were chosen.

src/widgets/buttons.ts
```diff
@@ -478,6 +488,42 @@ export function deleteButtonWithCheck (
   return deleteButtonElt
 }
 
+function getButtonStyle (options: ButtonWidgetOptions = { buttonType: ButtonType.Primary, filled: true }) {
+  // We need to accomadate for legacy code, which is why we have to allow buttonType and filled to be optional
+  let backgroundColor: (ButtonType | string | undefined) = options.buttonType
+  let fontColor: (ButtonType | string | undefined) = '#ffffff'
+  let borderColor: (ButtonType | string | undefined) = options.buttonType
+  let hoverBackgroundColor: (ButtonType | string | undefined) = `lighten(${options.buttonType}, 5%)`
```
SharonStrats: this lighten function here doesn't seem to work.  I could try to make this work or I could just set it to the color it would be if it were lightened by 5%...

src/widgets/buttons.ts
```diff
@@ -478,6 +488,42 @@ export function deleteButtonWithCheck (
...
+    borderColor = options.buttonType
+    hoverBackgroundColor = options.buttonType
+    hoverFontColor = backgroundColor
+  }
+
+  // See https://design.inrupt.com/atomic-core/?cat=Atoms#Buttons
+  return {
+    'background-color': `${backgroundColor}`,
+    color: `${fontColor}`,
+    'font-family': 'Raleway, Roboto, sans-serif',
```
SharonStrats: This font is not showing up because we need to add the font family to the library itself.  I will create a ticket.  should work once this is done.

Documentation/diagrams/solid-ui-modules.svg
```diff
@@ -0,0 +1,617 @@
...AlternateProcess" id="c93c3dc2c06f42b29203075146b145f1" transform="matrix(1,0,0,1,46,627)"><p:metadata><p:property name="box">160,38</p:property><p:property name="outputPin1Origin">80,0</p:property><p:property name="outputPin1">80,0</p:property><p:property name="outputPin2Origin">160,19</p:property><p:property name="outputPin2">160,19</p:property><p:property name="outputPin3Origin">80,38</p:property><p:property name="outputPin3">80,38</p:property><p:property name="outputPin4Origin">0,19</p:property><p:property name="outputPin4">0,19</p:property><p:property name="fillColor">#FFFFFFFF</p:property><p:property name="strokeColor">#666666FF</p:property><p:property name="strokeStyle">2|</p:property><p:property name="textContent">Buttons</p:property><p:property name="textFont">Helvetica|normal|normal|18px|none</p:property><p:property name="textColor">#000000FF</p:property><p:property name="textAlign">1,1</p:property><p:property name="optionText1"/><p:property name="optionText2"/><p:property name="optionText3"/><p:property name="optionText4"/><p:property name="optionTextFont">Helvetica|normal|normal|12px|none</p:property><p:property name="optionTextColor">#666666FF</p:property></p:metadata>
```
michielbdejong: Remove this from your PR

examples/buttons/index.html
```diff
@@ -1,31 +1,34 @@
 <!DOCTYPE HTML>
 <html lang="en">
-  <head>
-    <meta charset='UTF-8'>
-    <title>solid-ui UI.widgets.buttons examples page</title>
-    <script type="text/javascript" src="../../lib/webpack-bundle.js"></script>
-    <script>
-      const $ = document.querySelector.bind(document)
-      function showSource(widgetName) {
+
```
michielbdejong: why add a newline here?

examples/buttons/index.html
```diff
@@ -1,31 +1,34 @@
...
+<head>
```
michielbdejong: why change the indent here?

examples/buttons/index.html
```diff
@@ -90,23 +96,25 @@ <h2 id="button with icon"><a href="#button-with-icon">button with icon</a></h2>
...
 
- <h2 id="button-with-text"><a href="#button-with-text">button with text</a></h2>
+  <h2 id="button-with-text"><a href="#button-with-text">button with text</a></h2>
   <script id="script-button-with-text">
     window.addEventListener('DOMContentLoaded', (event) => {
       $('#div-button-with-text').appendChild(
         UI.widgets.button(document, undefined, 'test', () => {
           window.alert('clicked!')
-        })
+        }, { buttonType: UI.widgets.ButtonType.Secondary, filled: true })
```
michielbdejong: Please split this into 4 examples:
* primary, filled
* primary, not filled
* secondary, filled
* secondary, not filled

src/widgets/buttons.ts
michielbdejong: I thought the cancel and continue buttons use icons?
So yeah, we need to work out how we want to style icon-buttons.
In this PR it's fine if you look only at text buttons, not icon buttons.

src/widgets/buttons.ts
```diff
@@ -394,7 +404,7 @@ export function setImage (element: HTMLElement, thing: NamedNode) { // 20191230a
 // See eg http://stackoverflow.com/questions/980855/inputting-a-default-image
 function faviconOrDefault (dom: HTMLDocument, x: NamedNode) {
   var image = dom.createElement('img')
-  ;(image as any).style = style.iconStyle
+    ; (image as any).style = style.iconStyle
```
michielbdejong: what happened to the indent here, did your editor do this?

src/widgets/buttons.ts
```diff
@@ -478,6 +488,42 @@ export function deleteButtonWithCheck (
   return deleteButtonElt
 }
 
+function getButtonStyle (options: ButtonWidgetOptions = { buttonType: ButtonType.Primary, filled: true }) {
+  // We need to accomadate for legacy code, which is why we have to allow buttonType and filled to be optional
```
michielbdejong: I think it's not so weird to have a default, even regardless of legacy code

src/widgets/buttons.ts
```diff
@@ -478,6 +488,42 @@ export function deleteButtonWithCheck (
...
+  let backgroundColor: (ButtonType | string | undefined) = options.buttonType
```
michielbdejong: It's confusing to have two variables that mean the same thing but one is called type and the other is called color. Also having a variable of type `(ButtonType | string | undefined)` feels like it's an intermediate result; the end result you're working towards is a string, right? So maybe just use an if statement here, so:
```ts
let backgroundColor: string = PRIMARY_COLOR;
if (options.buttonType === Secondary) {
  backgroundColor = SECONDARY_COLOR;
}
```

src/widgets/buttons.ts
michielbdejong: yeah just set it to the colour.

src/widgets/buttons.ts
michielbdejong: yeah, this is fine. my only remark is it's best practice to name your options so that the default is negative.

So here you have filled vs. borderOnly.

If you want filled to be the default, you would call your option variable borderOnly (default: false).
If you want borderOnly to be the default, you would call your option variable filled (default: false).

That makes the code easier to reason about without having to deal with double negations in your head.

src/widgets/buttons.ts
michielbdejong: See #293.

src/widgets/buttons.ts
```diff
@@ -478,6 +488,42 @@ export function deleteButtonWithCheck (
...
+  let fontColor: (ButtonType | string | undefined) = '#ffffff'
+  let borderColor: (ButtonType | string | undefined) = options.buttonType
+  let hoverBackgroundColor: (ButtonType | string | undefined) = `lighten(${options.buttonType}, 5%)`
+  let hoverFontColor: (ButtonType | string | undefined) = fontColor
+  if (!options.filled) {
+    backgroundColor = '#ffffff'
+    fontColor = options.buttonType
+    borderColor = options.buttonType
+    hoverBackgroundColor = options.buttonType
+    hoverFontColor = backgroundColor
```
michielbdejong: This is confusing to read, you have five colours, but two are strings and three are types.

src/widgets/buttons.ts
```diff
@@ -487,7 +533,7 @@ export function deleteButtonWithCheck (
  *
  * @returns <dDomElement> - the button
  */
-export function button (dom: HTMLDocument, iconURI: string | undefined, text: string, handler: (event: any) => void) {
+export function button (dom: HTMLDocument, iconURI: string | undefined, text: string, handler: (event: any) => void, options: ButtonWidgetOptions = { buttonType: ButtonType.Primary, filled: true }) {
```
michielbdejong: See above, for readability, try to use option variables in such a way that false is the default.

src/widgets/buttons.ts
```diff
@@ -534,7 +566,7 @@ export function button (dom: HTMLDocument, iconURI: string | undefined, text: st
  * @returns <dDomElement> - the button
  */
 export function cancelButton (dom: HTMLDocument, handler: (event: any) => void) {
-  return button(dom, cancelIconURI, 'Cancel', handler)
+  return button(dom, cancelIconURI, 'Cancel', handler, {})
```
michielbdejong: if you make options optional, then you don't need to add `, {}` here.

src/widgets/buttons.ts
```diff
@@ -895,7 +927,7 @@ export function propertyTriage (kb: IndexedFormula): any {
   }
   possibleProperties.op = op
   possibleProperties.dp = dp
-  info(`propertyTriage: ${no} non-lit, ${nd} literal. ${nu} unknown.`)
+  info(`propertyTriage: ${no} non - lit, ${nd} literal.${nu} unknown.`)
```
michielbdejong: did you edit this on purpose?

src/widgets/buttons.ts
```diff
@@ -1130,13 +1162,13 @@ function twoLineTransaction (dom: HTMLDocument, x: NamedNode): HTMLElement {
...
   var box = dom.createElement('table')
   box.innerHTML = `
-    <tr>
-      <td colspan="2">${enc('payee')}</td>
-    </tr>
-    <tr>
-      <td>${enc('date').slice(0, 10)}</td>
-      <td style="text-align: right;">${enc('amount')}</td>
-    </tr>`
+      < tr >
```
michielbdejong: This looks wrong

src/widgets/buttons.ts
SharonStrats: there's both actually.  icons and then buttons.  I don't think buttons are in the examples.

src/widgets/buttons.ts
SharonStrats: yeah, it's my save.  I've tried to change it, but it just doesn't seem to work and the linter doesn't fix it.  is this okay or a problem?

src/widgets/buttons.ts
SharonStrats: I see what you are saying. I'll look a little further as I really thought it was better to use an enum.  I had to do undefined because of the legacy code, some may not pass in an option and it may be undefined.. basically typescript complained.

src/widgets/buttons.ts
SharonStrats: that sounds good.

src/widgets/buttons.ts
SharonStrats: yeah i could change to strings.. i didn't like it either, but there were so many variables that need to get set depending on what is selected.

src/widgets/buttons.ts
SharonStrats: good catch.  I changed it when I had made it required and didn't change it back.

src/widgets/buttons.ts
SharonStrats: no, my save :(

src/widgets/buttons.ts
SharonStrats: i didn't intentially change this. aww if you look below it's the spacing again.

examples/buttons/index.html
```diff
@@ -90,23 +96,70 @@ <h2 id="button with icon"><a href="#button-with-icon">button with icon</a></h2>
       )
     })
   </script>
-  <pre id="viewSource-button-with-icon"></pre><script>showSource('button-with-icon')</script>
+  <pre id="viewSource-button-with-icon"></pre>
+  <script>showSource('button-with-icon')</script>
   <div id="div-button-with-icon"></div>
 
- <h2 id="button-with-text"><a href="#button-with-text">button with text</a></h2>
-  <script id="script-button-with-text">
```
SharonStrats: I have added all the button options here in the examples, but the secondary isn't working right now due to the enum..  I will fix tomorrow.

src/widgets/buttons.ts
michielbdejong: See [button](https://solid.github.io/solid-ui/examples/buttons/#button), [cancelButton](https://solid.github.io/solid-ui/examples/buttons/#cancelButton), [continueButton](https://solid.github.io/solid-ui/examples/buttons/#continueButton)

src/widgets/buttons.ts
michielbdejong: sure, no problem

src/widgets/buttons.ts
michielbdejong: See https://github.com/solid/solid-ui/pull/289#discussion_r400060297 and https://github.com/solid/solid-ui/pull/289#discussion_r400066550

src/widgets/buttons.ts
michielbdejong: in this case it's not acceptable because it incorrectly changes the value of that string template.

michielbdejong:
@SharonStrats there was an inconsistency in the way you used defaults, for `color` you used constants and an `if` but for `hoverBackgroundColor` you used a conditional and hard-coded hex values: `(options.buttonColor === 'Primary') ? '#9f7dff' : '#37cde6'`. Also one of the hex color values was in upper case, I lower-cased that for consistency.

There was a bug when you specified needsBorder but not buttonColor, I fixed that by checking for `options.buttonColor === 'Secondary'` instead of `options.buttonColor === 'Primary'`.

I added comments to the four examples to make it clear which is which, and I removed the options that matched the defaults, to make it clearer that they can be omitted.

PR #291, opened 2020-03-27
Button with text and options for primary/secondary color, with/without border

michielbdejong:


examples/buttons/index.html
```diff
@@ -79,19 +79,32 @@ <h2 id="attachmentList"><a href="#attachmentList">attachmentList</a></h2>
   <pre id="viewSource-attachmentList"></pre><script>showSource('attachmentList')</script>
   <div id="div-attachmentList"></div>
 
- <h2 id="button"><a href="#button">button</a></h2>
-  <script id="script-button">
+ <h2 id="button with icon"><a href="#button-with-icon">button with icon</a></h2>
```
megoth: Needs dashes in id

test/unit/widgets/buttons.test.ts
```diff
@@ -105,7 +111,14 @@ describe('button', () => {
     const text = 'txt'
     const handler = () => {
     }
-    expect(button(dom, iconURI, text, handler)).toBeTruthy()
+    // FIXME: Not sure why this is needed, but
+    // JSS tries to do .insertBefore on a StyleSheet element,
+    // and that seems to fail in JSDOM (it works fine in the browser)
+    const domCopy = {
```
megoth: Probably want a helper for this quickly so that we can easily get a DOM-object with head for other tests as well.

test/unit/widgets/buttons.test.ts
```diff
@@ -105,7 +111,14 @@ describe('button', () => {
...
+      head: {},
+      createElement: dom.createElement.bind(dom)
+    }
+    expect(button(domCopy as unknown as HTMLDocument, iconURI, text, handler)).toBeTruthy()
```
megoth: You can do some more testing on this, e.g. is the property class set? You can also make a spy for the handler, and manually trigger it using `button.dispatchEvent(new Event('click'))`.

michielbdejong:
Merged Sharon's work into this PR

PR #294, opened 2020-04-01
Dependency map

timbl:
Add a map of internal dependencies and the tool to make it.


src/folders.js
```diff
@@ -7,7 +7,6 @@ import * as debug from './debug'
 const UI = {
   icons: require('./iconBase'),
   ns: require('./ns'),
-  pad: require('./'),
```
megoth, line 10: Is this change necessary in this PR?

src/matrix.js
```diff
@@ -23,7 +23,6 @@
 var UI = {
   icons: require('./iconBase'),
   ns: require('./ns'),
-  pad: require('./'),
```
megoth, line 26: Is this change necessary in this PR?

src/messageArea.js
```diff
@@ -4,7 +4,6 @@ var UI = {
   authn: require('./authn/authn'),
   icons: require('./iconBase'),
   ns: require('./ns'),
-  pad: require('./'),
```
megoth, line 7: Is this change necessary in this PR?

src/folders.js
timbl: I removed 3 unnecessary dependencies at the same time.  Can we let that go through in the same PR?

src/matrix.js
timbl: Not strictly but it is related and life is too short to separate them

src/messageArea.js
timbl: Not strictly but it is related and life is too short to separate them

michielbdejong (approved):

PR #295, opened 2020-04-02
New icons for @ # issue folder table ontology

timbl:
From the noun project as usual
For toolbar etc


package-lock.json
```diff
@@ -9167,7 +9167,7 @@
     },
```
michielbdejong, line 9167: I think this inclusion of package-lock.json changes was probably accidental? It seems unrelated to the topic of the PR (this commit was added after I did my review, sorry).

PR #298, opened 2020-04-06
Add header new files only

SharonStrats:
This only contains the new code, no modifications.  It has been tested with the examples/header test by hard coding the user to me and making sure it displays.  I then changed it back before pushing.


src/header/index.ts
```diff
@@ -0,0 +1,198 @@
...
+  if (fetcher) {
+    // Making sure that Profile is loaded before building menu
+    await fetcher.load(user)
+  }
+  // const outliner = getOutliner(document)
+
+  const loggedInMenuList = document.createElement('ul')
+  loggedInMenuList.classList.add('header-user-menu__list')
+  loggedInMenuList.appendChild(createUserMenuItem(createUserMenuLink('Show your profile', user.uri)))
+  /* const menuItems = await getMenuItems(outliner)
```
megoth, line 122: In general we don't want to leave commented out code left like this. But I reckon you're planning to remove this as part of continued work on this module?

src/header/index.ts
```diff
@@ -0,0 +1,198 @@
...
+  menuItems.forEach(item => {
```
megoth, line 123: I was thinking we should extract this list into something that is provided from mashlib? Part of next step you're working on?

SharonStrats: yeah this is the next time with the options.  it's actually in the add-header branch now.

src/header/index.ts
SharonStrats: I could just remove it now.  I think it's not a problem.

SharonStrats:
This is a reply to the message above about css, package.json and authentication.
- css: I thought I understood that I should do this after this gets merged?
- package.json happened because of the travis thing, should be good now.
- authentication.  Based on the way the code is written it looks at the session, I'm not sure how to make it authenticated and not authenticated on the same index.html file?

PR #303, opened 2020-04-26
Table hints - more control of the table widget

timbl:
You can select a default column to sort by, override labels, link functions and cell formatter for columns.

When pushed, had to push --no-validate as typescript testing  98 errors.  Not sue to these small changes, I assume rdflib changes.  @megoth any idea?


megoth:
@timbl I suspect you had npm linked to your local version of rdflib? When I pulled this down and checked locally, I was able to build without a problem.

src/table.js
```diff
@@ -455,10 +455,25 @@ module.exports = function renderTableViewPane (doc, options) {
       this.useCount += 1
     }
 
+    this.getHints = function () {
```
megoth, line 458: Could you add a test for this?

megoth:
And revert your changes to `package-lock.json` - no reason for why you need to change that file.

timbl:
(Yes, I had built it with the `editable` branch of https://github.com/linkeddata/rdflib.js/pull/412 ...)

PR #304, opened 2020-04-29
Possible fix after the types fix in rdflib

megoth:
This must  not be merged into master before we have the new types merged into rdflib - https://github.com/linkeddata/rdflib.js/pull/406


megoth:
Relies on changes in https://github.com/linkeddata/rdflib.js/pull/415

timbl:
The checks fail at the moment with
```
src/widgets/buttons.ts(922,15): error TS2339: Property 'toNT' does not exist on type 'Term'.
```

PR #315, opened 2020-07-13
Multiple bookmark files

michielbdejong:


test/unit/chat/bookmarks.test.ts
```diff
@@ -1,11 +1,18 @@
...
+
+const BOOK = Namespace('http://www.w3.org/2002/01/bookmark#')
 
 silenceDebugMessages()
 jest.mock('solid-auth-client')
 
 describe('findBookmarkDocument', () => {
+  afterEach(clearStore)
+
   it('exists', () => {
```
SharonStrats, line 16: I think we were changing this to something like ('it is exposed on the API'..  you can check out tabs.test.ts.

michielbdejong:
Thanks!

Issue #329, opened 2020-08-10
Run github.io from master instead of gh-pages


timbl:
I note the FIXME in

https://github.com/solid/solid-ui/edit/master/Documentation/developers/publish_README.md

Why not switch github.io to run from master instead of gh-pages?

angelo-v:
We would need to check-in the built files into master and mix it with the sources. I prefer keeping it in the gh-pages branch and automate the process as proposed in the readme. Will remove the storybook build from my PR as well.

timea-solid:
related to [[](url)](https://github.com/SolidOS/solid-ui/issues/568)

PR #332, opened 2020-08-26
Setup Storybook and button stories

angelo-v:
Basic storybook setup and stories for some button widgets. More can be added in separate PR, let's keep the original examples until everything is migrated.


storybook-static/iframe.html
```diff
@@ -0,0 +1,74 @@
+<!doctype html><html lang="en"><head><meta charset="utf-8"><title>Storybook</title><meta name="viewport" content="width=device-width,initial-scale=1"><base target="_parent"><style>:not(.sb-show-main) > .sb-main,
```
angelo-v, line 1: I commited the storybook static build as well, so we could base the github pages build on the main branch instead of gh-pages as proposed by @timbl in https://github.com/solid/solid-ui/issues/329

If we do not want to do that, I will remove the static files and add them to the gh-pages branch

.storybook/main.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+  stories: [
+    "../src/**/*.stories.mdx",
+    "../src/**/*.stories.@(js|jsx|ts|tsx)"
```
megoth, line 4: I guess these two lines could be shortened to `../src/**/*.stories.@(js|jsx|mdx|ts|tsx)`?

storybook-static/iframe.html
megoth: Might also be a case for checking out GitHub actions, to automatically build upon successful merges to master.

megoth (approved):
LGTM.

You might also want to check out the a11y addon, gives you some nice checks on accessability.

storybook-static/iframe.html
angelo-v: I removed it, let's automate the build instead

.storybook/main.js
angelo-v: those where generated, but indeed you are right

.travis.yml
```diff
@@ -13,12 +13,19 @@ script:
...
+  - provider: npm
+    # Do not throw away the updated package.json we generated in `before_deploy`:
+    skip_cleanup: true
+    email: '$NPM_EMAIL'
+    api_key: '$NPM_TOKEN'
+    # Note: do not deploy on pull request, because $TRAVIS_BRANCH will be the target branch.
+    tag: '$TRAVIS_BRANCH'
+    on:
+      all_branches: true
+  - provider: pages
```
angelo-v, line 25: @megoth I added gh-pages deployment using travis, since that is what we have in place anyway. I think someone with travis access (you?) needs to add a `$GITHUB_TOKEN`

michielbdejong (approved):

.travis.yml
megoth: You should have access, as you are administrator in this repo.

megoth: You don't have access to https://travis-ci.org/github/solid/solid-ui ? I think you should, as you are administrator in this repo.

angelo-v: You are right, wasn't aware of that

PR #336, opened 2020-09-01
remove built pages from gitignore before deployment

angelo-v:


angelo-v:
I guess it's fine if I merge this directly.

PR #338, opened 2020-09-13
migrate some more examples from examples/buttons to storybook

angelo-v:


michielbdejong:
I think to review this I should run `npx serve` or similar to simulate what http://solid.github.io/solid-ui/examples/storybook/ will look like once this is merged, right?

michielbdejong:
Ah, `npm run build-storybook`

angelo-v:
@michielbdejong or even simpler `npm run storybook` to startup a local dev server directly

PR #359, opened 2020-11-03
Move store to solid-logic

michielbdejong:
This branch moves the store into https://github.com/solid/solid-logic.
The contents that would be moved to that repo are still in `src/solid-logic-move-me.ts` to make debugging faster.
It also makes a start with moving some of the logic of `src/authn/authn.ts` to solid-logic.


michielbdejong:
CC @noeldemartin

src/logic.ts
```diff
@@ -0,0 +1,13 @@
+// This module of solid-ui has a main quadstore for the app to use
+//
+
+import * as debug from './debug'
+import { fetch } from 'solid-auth-client'
+// import { SolidLogic } from 'solid-logic'
+import { SolidLogic } from './solid-logic-move-me'
+
+export const solidLogicSingleton = new SolidLogic({ fetch })
```
angelo-v, line 9: I dislike the idea of having a global god singleton object containing all the logic. But if it is only a migration step into the idea of separating logic from ui I guess it is a valid start

angelo-v (commented):
I cannot overlook all the changes. If fell the need we can have a video call to go through it. Otherwise I trust you to act in all concience; merge at will.

src/solid-logic-move-me.ts
```diff
@@ -0,0 +1,181 @@
...
+      }) // callback
+    }) // promise
+  }
+
+  clearStore () {
+    this.store.statements.slice().forEach(this.store.remove.bind(this.store))
+  }
+}
+
+export class UnauthorizedError extends Error {
```
NoelDeMartin, line 134: I can't say much about the PR in general because I don't know the codebase, but I spotted a potential improvement here :).

I think it's better to define a custom base error (for example SolidLogicError) and extend everything from that. That way, it's not necessary to override the constructor in each error subclass.

Check out these files in soukai: [SoukaiError.ts](https://github.com/NoelDeMartin/soukai/blob/7f2575e20cbec8ddc9041da9436c2f69fa2b8696/src/errors/SoukaiError.ts) and [SoukaiError.test.ts](https://github.com/NoelDeMartin/soukai/blob/7f2575e20cbec8ddc9041da9436c2f69fa2b8696/src/errors/SoukaiError.test.ts).

src/logic.ts
michielbdejong: Well, it's a singleton instantiation of another code repo. In other code (e.g. Solid test-suite), I would probably instantiate multiple of these, but for the mashlib philosophy we want all data to go into one big searchable stack.
You are aware that the current code has a global god singleton object for the store, in `src/store.ts`, right? That's basically what this sits on on top of.

src/solid-logic-move-me.ts
michielbdejong: Nice! Thanks

src/logic.ts
angelo-v: I am fine with a single store as single source of truth. This does not mean, all logic has to be in one object or class. There are several concerns that should be split in the long run.

michielbdejong: Right, I see your point. So how about something like a factory pattern, where you have a single SolidLogic object around the store, but you use it to instantiate various objects of different classes (e.g. CurrentUser, AddressBookEntry, etc) that get an internal reference to the store when you instantiate them?

angelo-v: sounds good

angelo-v: It's fine for me if you want to merge first and do it in a separate PR

PR #362, opened 2020-11-09
Storybook table

michielbdejong:


michielbdejong:
https://github.com/search?q=org%3Asolid+renderTableViewPane&type=code
Hm, I can only find [one](https://github.com/solid/solid-panes/blob/06d473306d14723c6eba5af22a7b9c4d6dfad5ba/src/transaction/pane.js#L494) use of the table, and it's commented out and I find it hard to understand how to construct a query that will return results.
I'm seeing `Query done: 0 rows, 16ms` in the console all the time.

PR #367, opened 2020-11-30
Fix chat bug and decluttering tools in forms

timbl:
- Add ability to suppress form fields when the data is read-only. 
 - For security, add the fourth component to kb.each() etc .
 - Fix the bug that finite chat goes crazy loading data and displays nothing if it runs before the div is in place.


src/style.js
```diff
@@ -18,16 +18,26 @@ module.exports = {
...
+  personaBarStyle: 'width: 100%; height: 4em; background-color: #eee; vertical-align: middle;',
   // Login buttons
 
   signInButtonStyle: 'padding: 1em; border-radius:0.2em; margin: 2em; font-size: 100%;', // was 0.5em radius
   // Forms
 
+  heading1Style: 'font-size: 180%; font-weight: bold; color: #888888; padding: 0.5em;', // originall was brown now grey
+  heading2Style: 'font-size: 130%; font-weight: bold; color: #888888; padding: 0.4em;', // originall was brown now grey
+  heading3Style: 'font-size: 120%; font-weight: bold; color: #888888; padding: 0.3em;', // For eg in large forms or before a small form
+  heading4Style: 'font-size: 110%; font-weight: bold; color: #888888; padding: 0.2em;', // Lowest level used by default in smalll things
```
TallTed, line 31:
```suggestion
  heading1Style: 'font-size: 180%; font-weight: bold; color: #888888; padding: 0.5em;', // originally was brown; now grey
  heading2Style: 'font-size: 130%; font-weight: bold; color: #888888; padding: 0.4em;', // originally was brown; now grey
  heading3Style: 'font-size: 120%; font-weight: bold; color: #888888; padding: 0.3em;', // For example, in large forms or before a small form
  heading4Style: 'font-size: 110%; font-weight: bold; color: #888888; padding: 0.2em;', // Lowest level used by default in small things
```

src/style.js
```diff
@@ -18,16 +18,26 @@ module.exports = {
...
+
   formBorderColor: '#888888', // originall was brown now grey
   formHeadingColor: '#888888', // originall was brown now grey
   formHeadingStyle: 'font-size: 110%; font-weight: bold; color: #888888; padding: 0.2em;', // originall was brown now grey
```
TallTed, line 35:
```suggestion
  formBorderColor: '#888888', // originally was brown; now grey
  formHeadingColor: '#888888', // originally was brown; now grey
  formHeadingStyle: 'font-size: 110%; font-weight: bold; color: #888888; padding: 0.2em;', // originally was brown; now grey
```

src/widgets/buttons.ts
```diff
@@ -405,7 +405,7 @@ export function setImage (element: HTMLElement, thing: NamedNode) { // 20191230a
 
 // If a web page then a favicon with a fallback to
 // See eg http://stackoverflow.com/questions/980855/inputting-a-default-image
```
TallTed, line 407:
```suggestion
// If a web page, then a favicon, with a fallback to ???
// See, e.g., http://stackoverflow.com/questions/980855/inputting-a-default-image
```

src/widgets/forms/basic.ts
```diff
@@ -101,6 +101,9 @@ export function basicField (
     )
     return box
   }
+  // It can be cleaner to just remove empty feilds if you can't edit them anyway
```
TallTed, line 104:
```suggestion
  // It can be cleaner to just remove empty fields if you can't edit them anyway
```

TallTed (changes requested):
human-facing text edits in code comments

src/chat/infinite.js
```diff
@@ -787,6 +787,16 @@ export async function infiniteMessageArea (dom, kb, chatChannel, options) {
       !earliest.messageTable.initial &&
       earliest.messageTable.extendBackwards
     ) {
+      // If this has been called before the element is actually in the
+      // user's DOM tree, then this scrollTop check won't work -> loop forever
```
angelo-v, line 791: **question:** I don't get it from the comment yet: Is the "loop forever" fixed by these lines, or is there still an issue left?

src/widgets/forms.js
```diff
@@ -1183,7 +1197,7 @@ forms.makeDescription = function (
   }
   function saveChange (_e) {
     submit.disabled = true
-    submit.setAttribute('style', 'visibility: hidden; float: right;') // Keep UI clean
+    submit.setAttribute('style', ' : hidden; float: right;') // Keep UI clean
```
angelo-v, line 1200: **question**: Is this intentional? I wonder why `' : hidden;` is left without a preceding css property

src/chat/infinite.js
timbl: Its fixed by spotting that scrolHeight is zero, and putting off the load of the chat till 2 secs later

src/widgets/forms.js
timbl: Fixed in latest commit

PR #368, opened 2020-12-04
Type errors

angelo-v:
I tried to work arround the type errors that occured on master lately.

### DIFF
--- a/src/create/create.ts
+++ b/src/create/create.ts
@@ -88,6 +88,7 @@ export function newThingUI (
           _context => {
             const newPaneOptions: NewPaneOptions = Object.assign({
               newBase: newBase,
+              folder: options.folder || undefined,
               workspace: ws
             }, options)
             for (const opt in options) {
@@ -209,7 +210,7 @@ export function newThingUI (
   const mintingPanes = Object.values(thePanes).filter(pane => pane.mintNew)
   const mintingClassMap = mintingPanes.reduce((classMap, pane) => {
     if (pane.mintClass) {
-      classMap[pane.mintClass] = (classMap[pane.mintClass] || 0) + 1
+      classMap[pane.mintClass.uri] = (classMap[pane.mintClass.uri] || 0) + 1
     }
     return classMap
   }, {})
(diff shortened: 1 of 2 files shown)


src/create/create.ts
```diff
@@ -88,6 +88,7 @@ export function newThingUI (
           _context => {
             const newPaneOptions: NewPaneOptions = Object.assign({
               newBase: newBase,
+              folder: options.folder || undefined,
```
angelo-v, line 91: should folder be optional in NewPaneOptions?

src/create/create.ts
```diff
@@ -209,7 +210,7 @@ export function newThingUI (
   const mintingPanes = Object.values(thePanes).filter(pane => pane.mintNew)
   const mintingClassMap = mintingPanes.reduce((classMap, pane) => {
     if (pane.mintClass) {
-      classMap[pane.mintClass] = (classMap[pane.mintClass] || 0) + 1
+      classMap[pane.mintClass.uri] = (classMap[pane.mintClass.uri] || 0) + 1
```
angelo-v, line 213: Since this seems to be couting by class, using the URI is sufficient I guess

src/matrix/matrix.ts
```diff
@@ -71,7 +71,7 @@ export function matrixForQuery (
     header.setAttribute('style', 'padding: 0.3em;')
     header.textContent = utils.label(y1) // first approximation
     if (y1.termType === 'NamedNode') {
-      kb.fetcher.nowOrWhenFetched(y1.uri.split('#')[0], undefined, function (
+      kb.fetcher!.nowOrWhenFetched(y1.uri.split('#')[0], undefined, function (
```
angelo-v, line 74: Can we rely on that the fetcher is available. If so, can we reflect that in the type?

timbl: Yes.  There is a subclass of Store, LiveStore, (IIRC) which has fetcher and updater.

angelo-v: So, should the type in solid-logic changed to LiveStore @michielbdejong

PR #377, opened 2021-01-22
Add dot-generated SVG diagram of module dependencies - WIP

timbl:
This branch has an independent subdirectory where the various operations to grab dependencies and plot them happen.  the code worked by looking at lines with require r import but may need updating to recognize newer forms of import.


angelo-v:
We should try existing tools like [depcruiser](https://github.com/sverweij/dependency-cruiser), which work quite fine in my experience. The diagram generation can then be part of the build and visualizations can be deployed to gh-pages

theRealImy (approved):

PR #381, opened 2021-02-28
Solid logic update

angelo-v:
This PR upgrade to solid-logic 1.3.1, which requires an instance of solid-auth-client to be passed

I also did an npm audit fix


src/logic.ts
```diff
@@ -2,11 +2,10 @@
 //
 
 import * as debug from './debug'
-import { fetch } from 'solid-auth-client'
-// import { SolidLogic } from 'solid-logic'
+import auth from 'solid-auth-client'
 import { SolidLogic } from 'solid-logic'
 
-export const solidLogicSingleton = new SolidLogic({ fetch })
+export const solidLogicSingleton = new SolidLogic({ fetch: auth.fetch }, auth)
```
angelo-v, line 8: So, here we are passing the solid-auth-client instance, now

test/unit/acl/access-controller.test.ts
```diff
@@ -7,6 +7,10 @@ import { solidLogicSingleton } from '../../../src/logic'
 const store = solidLogicSingleton.store
 
 silenceDebugMessages()
+jest.mock('solid-auth-client', () => ({
+  currentSession: () => Promise.resolve(),
+  trackSession: () => null
```
angelo-v, line 12: `trackSession` is called from solid-logic now, so it has to be a function. I also added `currentSession`, to remove some error messages from the logs (which did not break the test, however)

michielbdejong (approved):
LGTM

PR #386, opened 2021-03-17
add a test for text button; set textContent instead of innerText

angelo-v:


michielbdejong:
I found https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext about this. Can you explain why we should change this?

michielbdejong (approved):
LGTM

angelo-v:
> I found https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext about this. Can you explain why we should change this?

The main cause I stumbled upon this is, that [testing-library](https://testing-library.com) was not able to find the button by text, so I tried to figure out what is wrong with the text.

By setting textContent we get a real text node into the dom like `<button>text</button>` while by setting innerText the dom only contains  `<button />`. If you revert back to innerText you can see the test I added fail. Also I found that textContent is standard, while innerText is non-standard introduced by IE.

PR #387, opened 2021-03-18
publish version to npm on main branch

angelo-v:
With this change, solid-ui will be automatically published to npm when a version is commited/merged to the main branch.

The other job, that I renamed to npm-publish-build, was already present and published everything to npm using the git slug. With this change we also publish a version without slug as latest npm package, but only when on main branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: .
 
-  npm:
+  npm-publish-build:
     needs: test
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


michielbdejong (approved):
LGTM
I would say, feel free to just edit things like this without peer review
What I'm wondering is how npm will react to `tag: latest`, but easiest way to find out is to try! :)

Does this also mean that it will automatically build whenever a git version is tagged? Isn't that what you really want to achieve?

NB 1: I think npm publish without further arguments will fail if the version number in the package.json was not incremented.

NB 2: I think 'latest' for npm means 'most recent' and not 'highest'

angelo-v:
Thanks for reviewing, Michiel. I hope you are not feeling annoyed by my review requests, feel free to ignore them, in case you do not have time.  I just don't want to change those things with at least someone taking note of it and give the possibility to veto

`latest` is just the default tag that is also applied when you do npm publish manually without specifiying something. I added it explicitly to be clear about what is happening.

When I understand `JS-DevTools/npm-publish@v1` correctly, it publishes a new version to npm whenever the version in the package.json changes, so it has nothing to do with git directly. But usually as we use `npm version` to increment the version a git tag will be created as well. If the version in package.json is untouched, the GitHub action will just do nothing, so there will be no fail.

PR #389, opened 2021-03-29
Modal for links

SharonStrats:
This is a modal. 
if you have an array of links [{ label: 'string', link: 'string'}, ... ]
You call createWindow(Dom, listofLinks) and it will return the modal. To display the modal by clicking a button add modal.display = 'block' to the onClickFunction. This has not been tested yet, due to some other technical issues.

I have also created a modalStyle file next to the modal. I think this is how we should do it as we go forward so that the styles are next to the file we are styling when possible. 

TODO: change the styling so that when the user hovers on li the li has a background color.


jeff-zucker (changes requested):
**[EDIT : ignore this, it was my setup]**

Haven't finished testing yet, but these changes allow it to compile without errors

Line 8 of modals.ts

  * change ../jss to ../lib/jss

Line 1 modalsStyles.ts

   * change ** to /*

angelo-v:
Cool. Could you please add unit tests and also stories to storybook?

timbl:
Is this a modal in the sense that it block all interaction for the user until the choice has been made?   If so then I would prefer that it is not. It should allow the user to work on other things while the choice has not been made yet.  Maybe it already does.

See https://www.w3.org/DesignIssues/UserInterface.html#modals
Also good to pass the function a DOM element near or under ot on top of which to make the pop-up so the choice doesn't get lost in the wrong part of the UI.

src/widgets/modals.ts
```diff
@@ -0,0 +1,95 @@
...
+// Two functions that need to be implemented to use the modal
+// When the user clicks the button, open the modal
+/* Click handler on the button to display it.
+btn.onclick = function() {
+  modal.style.display = "block";
+}
+
+// When the user clicks anywhere outside of the modal, close it
+Window click handler so that the modal will close
+even if the user doesn't click close
```
TallTed, line 24:
```suggestion
// Window click handler so that the modal will close
// even if the user doesn't click close.
```

src/widgets/modals.ts
```diff
@@ -0,0 +1,95 @@
...
+} from './modalsStyle'
+import { getClasses } from '../jss'
+
+export type ListItem = {
+  label: string,
+  link: string
+}
+
+// Two functions that need to be implemented to use the modal
+// When the user clicks the button, open the modal
```
TallTed, line 16:
```suggestion
// Two functions that need to be implemented to use the modal.
// When the user clicks the button, open the modal.
```

src/widgets/modals.ts
```diff
@@ -0,0 +1,95 @@
...
+/* Click handler on the button to display it.
+btn.onclick = function() {
+  modal.style.display = "block";
+}
+
+// When the user clicks anywhere outside of the modal, close it
```
TallTed, line 22:
```suggestion
// When the user clicks anywhere outside of the modal, close it.
```

TallTed (commented):
clarity changes to comments, and adding some `//` comment markers to lines without them in larger current `/* ... */` comment blocks (so if the `/* ... */` wrappers around the code are removed, these comment lines don't break the build)

src/widgets/modals.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+export const createWindow = (dom: HTMLDocument, listOfLinks: ListItem[], options: ModalWidgetStyleOptions) => {
+  const modal = createModal(dom, options)
+  const modalContent = createModalContent(dom)
+  const closeButton = createCloseButton(dom)
+  const ul = createUnOrderedList(dom, listOfLinks)
+  modalContent.appendChild(closeButton)
+  modalContent.appendChild(ul)
+  modal.appendChild(modalContent)
+}
```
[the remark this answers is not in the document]
SharonStrats: done

timea-solid:
What would help here is a screenshot of how the modal looks.

src/acl/access-groups.ts
```diff
@@ -264,7 +264,7 @@ export class AccessGroups {
         : 'No RDF type was detected for this URI.'
       const error =
         `Error: Failed to add access target: ${uri} is not a recognized ACL target type.` +
-        ` Expected one of: vcard:WebID, vcard:Group, foaf:Person, foaf:Agent, solid:AppProvider, solid:AppProviderClass, or recognized ACL classes.` +
+        ' Expected one of: vcard:WebID, vcard:Group, foaf:Person, foaf:Agent, solid:AppProvider, solid:AppProviderClass, or recognized ACL classes.' +
         ' Hint: try dropping a WebID profile URI, a vcard:Group URI, or a web app origin.' +
```
TallTed, line 268: It seems odd to combine (`+`) lines of backtick-wrapped text with lines of apostrophe-wrapped text. Given that line 267 is being changed from the former to the latter by this PR, it seems likely that line 266 (now backtick-wrapped) should be changed to match lines 267-268 (now apostrophe-wrapped).

SharonStrats:
I don't think we really need this anymore. if we do we can recreate in a web component.

PR #401, opened 2021-05-19
Timbl forms

timbl:
Merged publication work and dev work


timbl:
Adds checks for HTTP 200 status, (which is NOT checked by WebOperation), a few bugs.  Allows a a QueryParameters object so specify a suggested limit value

Documentation/forms-intro.html
```diff
@@ -545,6 +545,48 @@ <h4 class="class">Options</h4>
       </tr>
     </table>
 
+    <h4 class="class" id="Autocomplete">Autocomplete</h4>
```
angelo-v, line 548: **thought**: Would be cool to see the autocomplete in action in a storybook story as well.

src/widgets/forms/autocomplete/autocompleteField.ts
```diff
@@ -102,13 +103,13 @@ export function autocompleteField (
   const kb = store
   const formDoc = form.doc ? form.doc() : null // @@ if blank no way to know
 
-  const box = dom.createElement('tr')
+  const box = dom.createElement('div')
```
angelo-v, line 106: **praise**: Good to move away from table based layout

src/widgets/forms/autocomplete/autocompleteBar.ts
```diff
@@ -71,7 +70,7 @@ export async function renderAutocompleteControl (dom:HTMLDocument,
       creationArea.removeChild(decoratedAutocomplete)
       decoratedAutocomplete = undefined
     } else {
-      displayAutocomplete()
+      await displayAutocomplete()
```
angelo-v, line 73: **thought:** Souch a small word can make a big difference. Yet we still do not have any tests for this component, which worries me.

**suggestion**: Let's add automated tests for any components we change.

I have made some good experience with [testing-library](https://testing-library.com/docs/dom-testing-library/intro), since it helps a lot to test UI as a user experiences it. We need to do this to ensure no regressions come up and we already have a lot of Bugs in SolidOS which we need to get under control. Continuously improving and extending the test base is the only way to escape this situation in my experience.

src/widgets/forms.js
```diff
@@ -103,6 +102,15 @@ field[ns.ui('Form').uri] = field[ns.ui('Group').uri] =
       for (const x in already) already2[x] = 1
       already2[key] = 1
       const formDoc = form.doc ? form.doc() : null // @@ if blank no way to know
+      const weight0 = kb.any(form, ui('weight'), null, formDoc) // Say 0-3
```
angelo-v, line 105: **issue**: For me as a reviewer not directly involved in these changes it is very hard to crasp what all the changes in this file (form.js) are all about. The whole file is nearly 2.000 lines of code, It is even not obvious in which functions the changes are done. It is an anomymous function here, with no name, what is it's purpose? Corresponding tests could serve as specification of the desired behaviour, but there are none. Imho the file needs a whole lot of refactoring and testing before adding new features, otherwise we will be adding a whole lot of bugs as well.

src/widgets/forms.js
```diff
@@ -1208,32 +1243,35 @@ export function makeDescription (
     })
   }
 
-  const br = dom.createElement('br')
-  group.appendChild(br)
+  // const br = dom.createElement('br')
```
angelo-v, line 1246: **chore:** lets remove it if not needed anymore, commenting it out just leaves more clutter imho

src/widgets/forms/autocomplete/publicData.ts
```diff
@@ -290,6 +293,13 @@ export async function queryESCODataByName (filter: string, theClass:NamedNode, q
   return ESCOResultToBindings(json)
 }
 
+function fixWikidataJSON (str) {
+  const badness = str.indexOf('SPARQL-QUERY')
+  if (badness < 0) return str
+  debug.warn('  ### Fixing JSON with wikidata error code injection ' + str.slice(badness, badness + 200))
```
angelo-v, line 299: **question:** what is "badness" why is there a magic number of 200 added? What is it about?

**suggestion:** use named constants instead of magic numbers.

test/unit/widgets/forms/__snapshots__/basic.test.ts.snap
```diff
@@ -2,67 +2,66 @@
 
 exports[`basicField adds the basic field to the container, if provided 1`] = `
```
angelo-v, line 3: **thought:** So according to theses tests / snapshots only some styling changed and the layout switched to div-based. Looking at the changed code above this cannot be the whole truth.

**suggestion:** Cover all changes by tests, especially changes in behaviour.

angelo-v (changes requested):

src/style.js
```diff
@@ -5,18 +5,24 @@
 // These must all end with semicolon so they can be appended to.
 
 const formBorderColor = '#888888' // Mid-grey
+const lowProfileLinkColor = '#3B5998' // Grey-blue.   Eg for field labels linking to ontology
```
TallTed, line 8:
```suggestion
const lowProfileLinkColor = '#3B5998' // Grey-blue, e.g., for field labels linking to ontology
```

src/widgets/forms/autocomplete/autocompletePicker.ts
```diff
@@ -63,7 +63,7 @@ function assertNN (x):NamedNode {
 }
 */
 export function setVisible (element:HTMLElement, visible:boolean) {
-  element.style.visibility = visible ? 'visible' : 'collapse'
+  element.style.display = visible ? '' : 'none' // Do not use visibility, it holds the real esttate
```
TallTed, line 66:
```suggestion
  element.style.display = visible ? '' : 'none' // Do not use visibility, it holds the real estate
```

TallTed (changes requested):
typos

src/widgets/forms/autocomplete/autocompleteBar.ts
timbl: I thought that Typescript type checking would solve this, distinguishing between things which return `x` and `Promise<x>`

src/widgets/forms.js
timbl: I agree that forms.js needs dramatic restructuring, splitting it into main files, conversing to TS at the same time, and extracting named functions to reduce the complexity. As has been done with `forms/basic` for example, but applied to  the other fields.  Not on this branch.  (And look out for file-wide variable renaming we may want to do before we split it up.)

src/widgets/forms.js
timbl: agree

src/widgets/forms/autocomplete/autocompleteBar.ts
angelo-v: async functions always return Promise<x> (implicitly or explicitly), but this does not ensure whether the promise is awaited correctly. Anyway my suggestion is of a general nature: Let's at least start writing tests for issues we encounter, so that we will not encounter them again in the future. If there was an issue regarding a missing await we should have a test that turns red if the await is missing.

src/widgets/forms.js
angelo-v: I agree, we should not try to solve all problems at the same time. That is not what I suggest. However I highly recommend to add at least tests that cover the changes made to such a file. In my experience this is a viable strategy to get started with testing and the situation can improve continuously from there instead of waiting for the big future day where everything is refactored and tests are added.

src/widgets/forms/autocomplete/publicData.ts
timbl: "Badness is a flag that the returned entity is not valid JSON but is a horrible concatenation valid JSON and and timeout error message form a different level of the server system This is a massive kludge, because wikidata will return a huge list of bindings and then interrupt it with an error string starting "SPARQL-QUERY' when the server process times out.  The JSON parser which comes by default with the JS world throws a syntax error and so does not return any data, which is ironic because in these cases the number of bindings returned before the timeout is typically very large.  How best to deal with this? Link a wikidata issue and a solid-ui issue?  
The 200 is just an arbitrary length of the error message received passed onto the debug.warn stream .... there is normally a lot more, including a Java stack IIRC.

angelo-v: ok, ugly thing... so, what we could do about it is at least leave this helpful explanation you gave here as a comment in the code so that people can find it. Or yes, create issues on wikidata/and or solid-ui and leave a link to the issue as a comment in the code.

test/unit/widgets/forms/__snapshots__/basic.test.ts.snap
timbl: Well there were no tests for the autocomplete field... working on that.

Documentation/forms-intro.html
timbl: I have tests now for the autocomplete

timbl: they tell stories..

src/widgets/forms/autocomplete/autocompleteBar.ts
timbl: Yes.  Makes sense a a good way to work.    We can't be the only people thinking this ... maybe it will turn up in a new version of tsc or eslint.

src/widgets/forms.js
timbl: Yes.

timbl: [timbl-forms 2ae2d0d] remove commented-out code lines

src/widgets/forms/autocomplete/publicData.ts
timbl: Re-written with diff variable name.   Included a test using the bad data to make sure it works.

src/widgets/forms.js
timbl: I have added tests for the autocomplete form field, with two paths to 'accept' and one cancel path.   But only testing the case of there being existing data, not when there is no existing data, or testing all combinations of parameters to the form.

src/widgets/forms/autocomplete/publicData.ts
timbl: Re-written and test added to make sure it works.

src/widgets/forms/autocomplete/autocompleteField.ts
timbl: 👍

test/unit/widgets/forms/__snapshots__/basic.test.ts.snap
timbl: Tests done now

PR #402, opened 2021-05-28
Improve labels

angelo-v:
- adds support to render as:name and schema:name as label
- refactors the label utility function and migrates it to typescript
- adds unit test for existing and new functionality


src/utils/label.ts
```diff
@@ -0,0 +1,112 @@
...
+        s =
+          s.slice(0, i) +
+          String.fromCharCode(parseInt(s.slice(i + 1, i + 3), 16)) +
+          s.slice(i + 3)
+      }
+    }
+  }
+  if (s.slice(-5) === '#this') s = s.slice(0, -5)
+  else if (s.slice(-3) === '#me') s = s.slice(0, -3)
+
```
timbl, line 81: How about:
```
if s.endsWith('/profile/card') {
return s.split('/')[2].split('.')[0] // Use domain name
}
```

timbl (changes requested):
Happy for this request for changes to be ignored too just make the change a cleanup change but I thnk recognizing profile/card as a special case is straight forward and would improve dramatically for example  data view of people's friends lists, which are currently 'card, card card, card'!

angelo-v:
That's a great idea and should be easy to add, so I will include it

src/utils/label.ts
angelo-v: I kept the slicing for now, and extraced it to a small function at least. I am not totally happy with it but don't have time right now. If you are fine with it feel free to merge, otherwise I will try to find some time for further refactoring later

timbl: Lets go with this

timbl (approved):

PR #417, opened 2021-06-24
remove alert and confirm from sharing-pane

bourgeoa:
- Since `pod owner` with `Control` has been added to NSS there is no need to limit user options in building ACL's 
- removed false comment in file.acl that the acl applies to the included folders

Tested on a local multi-user NSS with pod owner.


bourgeoa:
issue #417

PR #420, opened 2021-08-05
[WIP] Auth Upgrade

jaxoncreed:


jaxoncreed:
Tests fail because of the dependence on Solid-Logic

src/header/index.ts
```diff
@@ -76,14 +76,17 @@ export async function initHeader (store: IndexedFormula, options?: HeaderOptions
...
 }
 /**
  * @ignore exporting this only for the unit test
  */
 export function rebuildHeader (header: HTMLElement, store: IndexedFormula, pod: NamedNode, options?: HeaderOptions) {
-  return async (session: SolidSession | null) => {
-    const user = session ? sym(session.webId) : null
+  return async () => {
+    const sessionInfo = authSession.info
+    const user = sessionInfo.webId ? sym(sessionInfo.webId) : null
```
SharonStrats, line 89: I think here is where @jeff-zucker is saying we should use currentUser() instead of sessionInfo.webId as described in https://github.com/solid/solidos/blob/main/documentation/Using_Private_Resources_In_SolidOS.md

SharonStrats (changes requested):
I'm going to go through all individually and then test all of SolidOS locally.  It would be good if we could have more tests.  I will see what I can do.

src/header/index.ts
jeff-zucker: @SharonStrats - yes that is one of the places.  Also in the dashboardPane runBuildPage and also in other places.  Currently we have these predicates - currentUser, defaultTestUser, offlineID, checkUser - all of which return a webid and are used in various places.  This is on top of the places that authSession.webId is called directly.  With things like the dashboard  it's clear that we don't need to check for an actual session to display the dashboard - it can be on a private no-auth page.  Other places in the code are harder to figure whether we need to be checking for the actual session rather than a "acting as logged in".  The stuff in Solid Logic in is particularly hard for me to follow.

jeff-zucker: @SharonStrats - could you take a look at https://github.com/solid/solid-panes/pull/300/commits/c9f33b4b9c60ef1823724e66448a15b28784bb30#diff-c5076b653b08366bc6f16085ac8d0738cba36540a3b06b5f0fb913cdb03e664f - I tried to make the change but something is not right.

PR #422, opened 2021-08-16
Add ResearchOrganization map to wikidata class and fix bug

timbl:


SharonStrats (commented):
The changes look good.  I just wanted to double check the change in package-lock.json? I couldn't tell by the other change what happened with file-uri-to-path.

PR #423, opened 2021-08-24
Data-Kitchen support in authn.ts

jeff-zucker:
Checks for window.SolidAppContext, if set AND visiting a page that we don't want to log into, sets me = SolidAppContext.webid.


jaxoncreed:
Is AppContext a Data-Kitchen specific thing or was that already in there?

PR #425, opened 2021-08-25
fixed ReferenceError: TextEncoder is not defined

theRealImy:
I found an elegant way to fix the TextEncoder is not defined issue.


bourgeoa:
This is working perfectly.

Issue #152, opened 2019-10-28
Include Link to User Guide


kevin-inrupt:
## User Story
As a User, I want to access the Data Browser user guide from the application, so I can get help on how to use the Data Browser.

## Acceptance Criteria
1. Open the Data Browser.
2. TBD...

## Dev Notes
- A couple of potential implementation options are:
  1. Include a link to the user guide in the top-right dropdown menu. This would require the menu to be displayed even if the user is not-authenticated.
  2. Include a ? button in the header.

PR #433, opened 2021-10-01
a PR related to #152

theRealImy:
Solid-ui has a header component which is not used in production now but the plan was to use it. 
I wanted to keep the code up to date reason why I added it to the UI too.
Again: this change DOES NOT change anything in production but is still relevant to be updated.


jeff-zucker:
This is great, thanks!  I'm trying to figure out why the header is not showing up in Data-Kitchen.  I thought I had used currentUser() in all needed places, but I guess I missed one. Are you saying that the code for the header is in two places?  Where should I look?

theRealImy:
> @theRealImy - This is great, thanks! I'm trying to figure out why the header is not showing up in Data-Kitchen. I thought I had used currentUser() in all needed places, but I guess I missed one. Are you saying that the code for the header is in two places? Where should I look?

The code that displays the header in production is the one in mashlib: https://github.com/solid/mashlib/pull/125
The header code in solid-ui is only for the storybook.

Issue #439, opened 2021-10-17
Adding new "thing" on Your Stuff (homePane)


SharonStrats:
When I create a new account and go to "Your Stuff" to create something for the first time, I recieve this error.

![](https://files.gitter.im/5ee917dbd73408ce4fe71415/RWfP/Screen-Shot-2021-10-17-at-9.05.49-pm.png)

@timbl I want to make sure what you said you wanted to do with workspaces? Should I create a workspace for the user if they don't exist or should I just delete the fact it's checking for one? Also, can you tell me what file.id() is trying to do by chance? 

This code is in src/authn/authn.ts line 1362.

SharonStrats:
Comment from Tim on gitter The message “you don’t seem to have any workspaces. You have one storage spaces” iis not rally an error, it is informational, and so should have a white background.
However, “`file.id is not a function” is clearly an error.

PR #440, opened 2021-10-31
fix(authn): workspace message white and file id error

SharonStrats:
To fix https://github.com/solid/solid-ui/issues/439
Screen Shot 2021-10-31 at 4 57 11 pm


theRealImy (approved):
checked locally and it works.

Issue #441, opened 2021-10-31
Can't create first chat file of the day with only append permission


timbl:
If you happen to the person who makes the first message of the day, and you are a user who has append permission only to the chat space, then you can't create that first  file because that requires Write not Append.  
(This is not a problem if the message by an owner who created the chat)

timbl:
Can e in fact solve this by doing a PATCH ... INSERT instead of a PUT to create each chat file -- or do the WAC rules require you to have Write access for that too?

timbl:
Should we check whether the PUT should in fact have been rejected for those with Append perms?

timbl:
See https://gitter.im/solid/solidos?at=617cea2798c13e755005f8c8 and following

timbl:
```
   createMessageTable for  Fri Oct 29 2021 01:00:00 GMT+0100 (British Summer Time)
mashlib.min.js:8 @@@ SERVER_MKDIRP_BUG: Should only happen once: create chat file: <https://drive.verborgh.org/public/2021/chat/2021/10/31/chat.ttl>
mashlib.min.js:8 createIfNotExists: doc does NOT exist, will create... <https://drive.verborgh.org/public/2021/chat/2021/10/31/chat.ttl>
/public/2021/chat/2021/10/31/chat.ttl:1 Failed to load resource: the server responded with a status of 403 ()
mashlib.min.js:8 createIfNotExists doc FAILED: <https://drive.verborgh.org/public/2021/chat/2021/10/31/chat.ttl>: Error: Web error: 403 on PUT of <https://drive.verborgh.org/public/2021/chat/2021/10/31/chat.ttl>
mashlib.min.js:67 Uncaught (in promise) Error: Web error: 403 on PUT of <https://drive.verborgh.org/public/2021/chat/2021/10/31/chat.ttl>
    at mashlib.min.js:67
```

timbl:
Note the `@@@ SERVER_MKDIRP_BUG` sound like different client code being temporarily put in to compensate. for non-standard Solid servers which do not generate all the intervening directories in the path for a a patch.

timbl:
Maybe setting `SERVER_MKDIRP_BUG` to `false` will fix the problem by replacing the PUT with a PATCH

RubenVerborgh:
> Can e in fact solve this by doing a PATCH ... INSERT instead of a PUT to create each chat file -- or do the WAC rules require you to have Write access for that too?

[...]

PR #442, opened 2021-11-01
Try to address https://github.com/solid/solid-ui/issues/441

timbl:
This removed the PUT and leaves only a PATCH which should work.

### DIFF
--- a/.web_base
+++ b/.web_base
@@ -1,0 +1,1 @@
+https://github.com/solid/solid-ui/tree/main
--- a/src/chat/infinite.js
+++ b/src/chat/infinite.js
@@ -22,7 +22,7 @@ const UI = { authn, icons, ns, media, pad, $rdf, store, style, utils, widgets }
 
 /* global alert */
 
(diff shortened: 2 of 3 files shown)


timbl:
See https://github.com/solid/solid-ui/issues/441

.web_base
```diff
@@ -0,0 +1 @@
+https://github.com/solid/solid-ui/tree/main
```
RubenVerborgh, line 1: Do we need this document?

RubenVerborgh (approved):
Tested with the CSS Mashlib recipe and confirmed to be working.

src/chat/infinite.js
```diff
@@ -209,6 +209,7 @@ export async function infiniteMessageArea (dom, kb, chatChannel, options) {
           chatDocument
         )
         await createIfNotExists(chatDocument)
+        // Otherwise, create the documentif necessary with a PATCH:
```
TallTed, line 212:
```suggestion
        // Otherwise, create the document, if necessary, with a PATCH:
```

Issue #443, opened 2021-11-02
Does not suggest the current host as an OIDC issuer


RubenVerborgh:
When opening Mashlib from my server instance, it only suggests a canned list of OIDC issuers:

<img width="1005" alt="image" src="https://user-images.githubusercontent.com/675313/139861180-65e5cf87-e8fb-4f28-8cc5-ef1563527841.png">

However, in such cases, it would be very helpful to also list the current server as an option (in this case `drive.verborgh.org`), given that Mashlib is likely running on a Solid server.

PR #444, opened 2021-11-02
Suggest the current host as an OIDC issuer

RubenVerborgh:
image

Closes https://github.com/solid/solid-ui/issues/443


bourgeoa (approved):
LGTM Nice idea

RubenVerborgh:
Thanks, @bourgeoa! I have another idea related to this, which I might PR shortly.

PR #449, opened 2021-11-18
use acl:default when writing ACL, not the depcrecated acl:defaultForNew

angelo-v:
use of defaultForNew is deprecated and caused trouble on CSS


bourgeoa (approved):
- Good catch there is an other one in NSS https://github.com/solid/node-solid-server/blob/main/default-templates/new-account/.well-known/.acl 
- wanted to change options.defaultForNew to options.default but not sure where it is used
- resolved the failing build with solid-logic@1.3.9

angelo-v:
> wanted to change options.defaultForNew to options.default but not sure where it is used

I thought about this as well and discarded it for the same reason (at least for now)

> resolved the failing build with solid-logic@1.3.9

great, thank!

Issue #453, opened 2021-11-21
Selector: can't mint new with no subform. - New bogus message


timbl:
This message "Selector: can't mint new with no subform." appears in the UI of forms which used to wok.
The subform does not have to be set directly, it can be found by setting "from class" and then that class having a creation form.

timbl:
The code complains is a Choice has no subform, but although it is useful to have one, it is not essential, as a subform canbe found from the class of the thing to be created, wand that class is always provided.

PR #456, opened 2021-11-22
Remove bogus error condition

timbl:
To fix https://github.com/solid/solid-ui/issues/453

### DIFF
--- a/src/widgets/forms.js
+++ b/src/widgets/forms.js
@@ -1295,11 +1295,8 @@ export function makeSelectForOptions (
         predicate +
         '.'
     )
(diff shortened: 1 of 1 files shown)


src/widgets/forms.js
```diff
@@ -1295,10 +1295,11 @@ export function makeSelectForOptions (
         predicate +
         '.'
     )
-  }
+  }/* See https://github.com/solid/solid-ui/issues/453
```
angelo-v, line 1298: **suggestion** remove it, instead of commenting it out. Reference the issue in the PR and the commit message to document why

timbl: Agree

PR #461, opened 2021-12-21
Move header and footer out of mashlib

theRealImy:
This work builds on top of previous work done by Michiel and Sharon. 
I changes some type definitions, added the footer to it, and fixed tests and lint issues. 
Remark: this change will not affect anything in production. Only changed on mashlib do!
Initial issue: https://github.com/solid/mashlib/issues/107


src/footer/index.ts
```diff
@@ -0,0 +1,90 @@
+/*
+    This file was copied from mashlib/src/global/footer.ts file. It is modified to
+    work in solid-ui by adjusting where imported functions are found.
+ */
+import { IndexedFormula, NamedNode } from 'rdflib'
+import { authSession, currentUser } from '../authn/authn'
+import { addStyleClassToElement, getName, getPod, getPodOwner } from '../utils/headerFooterHelpers'
+
+const DEFAULT_SOLID_PROJECT_URL = 'https://solidproject.org'
```
SharonStrats, line 9: suggestion: I'm wondering since these were also in solid-panes, perhaps we could just export them from solid-ui and include them as well in solid-panes.

src/footer/styleMap.ts
```diff
@@ -0,0 +1,7 @@
+export const styleMap = {
+  footer: {
```
SharonStrats, line 2: suggestion: you can put this in our style.js file.  Most styles are there. I do believe we could break this up though, but in a style folder with separate files for each component's styles and then export them all from an index...

src/header/index.ts
```diff
@@ -4,88 +4,99 @@
...
   banner.appendChild(podLink)
 
   const leftSideOfHeader = document.createElement('div')
   addStyleClassToElement(leftSideOfHeader, ['header-banner__right-menu'])
   leftSideOfHeader.appendChild(userMenu)
-  leftSideOfHeader.appendChild(helpMenu)
+
+  if (options && options.helpMenuList) {
+    const helpMenu = createHelpMenu(options, options.helpMenuList)
+    leftSideOfHeader.appendChild(helpMenu as HTMLDivElement)
```
SharonStrats, line 98: praise: love the descriptive name!

SharonStrats (approved):
Your code looks great.  I just had a few suggestions.  I haven't brought them all down together to test it, but I'm wondering if maybe @bourgeoa could put them up on test for us, so we can all have a go since it is touching a bit. Love how you separated things and cleaned it up! Amazing work!

src/footer/index.ts
theRealImy: They are the same for now but the logic allows you to customize them as you need.
Example: 
- someone used the header solid-ui elements for their app -> there are default settings BUT theoretically, in their own app they can define other links or menu names. 
- solid-panes is just a customizable version of the solid-ui  header elements/widgets

src/footer/styleMap.ts
theRealImy: Let's tackle this when with the style/UI refactoring work.

PR #462, opened 2021-12-27
cleaned out LiveStore from pane-registry

theRealImy:
In this PR I clean out the LiveStore from pane-registry and used the solid-logic one instead. This is part of a bigger refactoring to get LiveStore streamlined.


bourgeoa (approved):
LGTM

PR #472, opened 2022-01-30
Protect against multiple participation objects; and un-writeable doc.

timbl:
Aimed at fixing https://github.com/solid/chat-pane/issues/71


src/participation.ts
```diff
@@ -1,6 +1,6 @@
 /* Manage a UI for the particpation of a person in any thing
 */
-
+import * as debug from '../debug'
```
SharonStrats, line 3:
```suggestion
import * as debug from '../src/debug'
```

src/participation.ts
```diff
@@ -149,6 +157,10 @@ export function recordParticipation (subject: NamedNode, padDoc: NamedNode, refr
     // If I am not already recorded
     return parps[0] // returns the particpation object
   } else {
+    if (!store.updater.editable(padDoc)) {
+      debug.info('Not recording participation, as no write acesss as ' + me + ' to ' + padDoc)
```
SharonStrats, line 161:
```suggestion
      debug.log('Not recording participation, as no write acesss as ' + me + ' to ' + padDoc)
```

src/participation.ts
```diff
@@ -19,7 +19,7 @@ type ParticipationOptions = {
 class ParticipationTableElement extends HTMLTableElement {
```
SharonStrats, line 19: I had an idea to try of a Map, we could add 

type Candidate = Map<string,NamedNode>

src/participation.ts
```diff
@@ -87,14 +87,25 @@ export function renderPartipants (dom: HTMLDocument, table: ParticipationTableEl
 export function participationObject (subject: NamedNode, padDoc: NamedNode, me: NamedNode) {
```
SharonStrats, line 87: let candidate: Map<string,NamedNode>

src/participation.ts
```diff
@@ -87,14 +87,25 @@ export function renderPartipants (dom: HTMLDocument, table: ParticipationTableEl
...
 
-    const parps = kb.each(subject, ns.wf('participation')).filter(function (pn) {
-      return kb.holds(pn, ns.wf('participant'), me)
+    const parps = store.each(subject, ns.wf('participation')).filter(function (pn) {
+      return store.holds(pn, ns.wf('participant'), me)
     })
-    if (parps.length > 1) {
-      throw new Error('Multiple records of your participation')
+    if (parps.length > 1) { // This can happen. https://github.com/solid/chat-pane/issues/71
+      const candidates = []
```
SharonStrats, line 97:
```suggestion
      const candidates: Candidate[] = []
```

src/participation.ts
```diff
@@ -87,14 +87,25 @@ export function renderPartipants (dom: HTMLDocument, table: ParticipationTableEl
...
+      for (const participation of parps) {
+        const date = store.anyValue(participation, ns.cal('dtstart'))
+        if (date) {
+          candidates.push([date, participation])
```
SharonStrats, line 101:
```suggestion
          candidate = candidate.set(date, participation)
          candidates.push(candidate)
```

SharonStrats (approved):
I tried to help with the type errors as I know you are busy.  I did not test though to see if the map would work with the sort. It does solve the Typescript errors though.  My system isn't setup perfectly right now, but I can try it out tomorrow.  I will check to see if you already fixed it tomorrow and if not I'll see if I can test it locally.

src/participation.ts
timbl: I used an array so as to be able to sort by date.. surely a set or map doesn't have an order and so wont sort.

SharonStrats: I was worried about that :(. but I tried to do some other typing but couldn't get it to work.  since it was pushing an array on an array.  I'll do more research.

src/participation.ts
SharonStrats: We can do this instead 
type CandidateEntry = string | NamedNode
type CanidateEntries = CandidateEntry[]

src/participation.ts
SharonStrats: then
 let candidate: CanidateEntries

bourgeoa:
Resolve conversations to hope make the tests run. They are blocked for one day now.

PR #479, opened 2022-03-05
adjust the styling lines for the header

SharonStrats:
Create a more subtle menu
Screen Shot 2022-03-06 at 7 22 01 am


theRealImy:
I really like softer line.
P.s I recently changes something in mashlib regarding css and I notices a small change. The header was always meant to be in a shadow box actually... the size of it was over the window size. So not it looks like:
image
Do you think this is a problem on your style?

SharonStrats:
> I really like softer line. P.s I recently changes something in mashlib regarding css and I notices a small change. The header was always meant to be in a shadow box actually... the size of it was over the window size. So not it looks like: image Do you think this is a problem on your style?

Yeah, I changed it on purpose.  Do you not like it? I can put it back.

theRealImy:
I am fine with your changes.

SharonStrats:
no longer needed

PR #483, opened 2022-03-23
Tabs bug fix

timbl:
This fixes a big with the default tab in tabs, and some discovery issues


timbl:
Also moves to async and renames IndexedFormula to Store

src/acl/access-groups.ts
```diff
@@ -58,13 +58,13 @@ export class AccessGroups {
   public aclMap: AgentMapMap
   private readonly addAgentButton: AddAgentButtons
   private readonly rootElement: HTMLElement
-  private _store: LiveStore
+  private _store: Store // @@ was LiveStore but doe not need to ne connected to web
```
TallTed, line 61:
```suggestion
  private _store: Store // @@ was LiveStore but does not need to be connected to web
```

PR #485, opened 2022-03-29
Extension false

bourgeoa:


theRealImy (commented):
Good catch on the fix!
Regarding all the dependency version updates -> I would separate the PR in 2: one for the fix and one for update dep.

PR #491, opened 2022-04-02
load container representation

bourgeoa:
this only should only solve part of the issue see https://github.com/solid/solidos/issues/124


theRealImy (approved):
Looks good to me.

PR #501, opened 2022-05-09
improved ui:Choice code

timea-solid:
Here my code improvements for ui:Choice. Among others:
-> when one adds a new Node it will only be created if at least one of its fields is edited successfully. 
-> subForm is refresh -> it is deleted and rendered again (probably can be solved more elegant with using refresh() -> next iteration)
-> made sure it still works with all examples we have (when BlankNodes or missing labels)

### DIFF
--- a/Documentation/form-examples/structures3.html
+++ b/Documentation/form-examples/structures3.html
@@ -1,0 +1,69 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+
+<html xmlns="http://www.w3.org/1999/xhtml">
+
+<head>
+    <meta name="generator"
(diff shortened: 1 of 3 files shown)


Documentation/form-examples/structures3.html
```diff
@@ -0,0 +1,69 @@
...
+<head>
+    <meta name="generator"
+        content="HTML Tidy for Mac OS X (vers 31 October 2006 - Apple Inc. build 15.17), see www.w3.org" />
+    <meta http-equiv="Content-Type" content="text/html; charset=us-ascii" />
+
+    <title>Form structures 4 - Choice</title>
+    <link type="text/css" rel="stylesheet" href="../style/demo-style.css" />
+
+    <!-- used for production ready documentation -->
+    <!-- note: one does not need whole mashlib, solid-ui is sufficient also -->
```
TallTed, line 14:
```suggestion
    <!-- used for production-ready documentation -->
    <!-- note: one does not need whole mashlib; solid-ui is sufficient -->
```

timbl (approved):
As a cursory scan

PR #511, opened 2022-07-22
Form choice fix

timea-solid:
Hack to fix a rdflib problem in forms ui:Choice


src/widgets/forms.js
```diff
@@ -1978,3 +1978,67 @@ function containsObject (obj, list) {
 
   return false
 }
+
+// This functions replaces the findMembersNT (thisClass) from rdflib until we fix: https://github.com/linkeddata/rdflib.js/issues/565
+/**
+ * For thisClass or any subclass, anything which has it is its type
+ * or is the object of something which has the type as its range, or subject
+ * of something which has the type as its domain
+ * We don't bother doing subproperty (yet?)as it doesn't seeem to be used
```
TallTed, line 1987: @timea-solid 
```suggestion
 * For anything which has thisClass (or any subclass) as its type,
 * or is the object of something which has thisClass (or any subclass) as its range, 
 * or subject of something which has thisClass (or any subclass) as its domain
 * We don't bother doing subproperty (yet?) as it doesn't seem to be used
```

TallTed (commented):
Merged too quickly after creation of PR!

src/widgets/forms.js
timea-solid: Ups.
I did the fix on main directly. 👍🏻

PR #512, opened 2022-08-14
Support login from Data Kitchen

jeff-zucker:
The box.refresh method is one place where we can NOT use authn.currentUser() because what we want to know is not just is the user authorized to be here but are they actually logged in.  So we need authn.authSession.info here.  Otherwise when we are in Data-Kitchen we appear to be logged in to the local host  (so we can access it all) but we aren't actually logged in to it.  However we want the option to login to other places so box.refresh needs to check whether we are actually logged in somewhere else and if not give us the issuer popup.

### DIFF
--- a/src/login/login.ts
+++ b/src/login/login.ts
@@ -591,6 +591,7 @@ export function renderSignInPopup (dom: HTMLDocument) {
       border-radius: 4px;
       min-width: 400px;
       padding: 10px;
+      z-index : 10;
     `
   )
   issuerPopup.appendChild(issuerPopupBox)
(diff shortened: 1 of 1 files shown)


jeff-zucker:
I am changing this to a bug.  It used to work fine.  Someone changed it, it no longer works, this is a regression.

bourgeoa:
@jeff-zucker CI still not working the issue is with 
L790 sym not defined

jeff-zucker:
> @jeff-zucker CI still not working the issue is with L790 sym not defined

@bourgeoa - sorry, fixed now, CI passes.

jeff-zucker:
@timea-solid and @bourgeoa - I updated the documentation to more clearly explain why this PR is necessary : https://github.com/SolidOS/solidos/blob/main/documentation/Using_Private_Resources_In_SolidOS.md

timea-solid (approved):

PR #537, opened 2023-03-24
data source option

bourgeoa:
Added a check on the following triple : 
```
 form ui:options "search-full-store"
```
different solution can also be used like .
```
form ui:search-full-store true
```
Or anything along this idea.

### DIFF
--- a/src/widgets/forms.js
+++ b/src/widgets/forms.js
@@ -783,11 +783,11 @@ field[ns.ui('Choice').uri] = function (
 
   const opts = { form, subForm, disambiguate: false }
 
-  function getSelectorOptions () {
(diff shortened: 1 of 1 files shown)


bourgeoa:
I just saw that `ui:Options` is already used. Then may be that lower case `options` is a bad idea.
I am not sure of the `ui` predicate either.

I found the `rdf:Property` in such cas we can use : 
```
form rdf:Property "search-full-store"
```

timea-solid (approved):
I think it is fine. We shall test on test server.

bourgeoa:
following SolisOS today's meeting

timea-solid (approved):

PR #542, opened 2023-04-10
Apply patch

SharonStrats:
checking


package.json
```diff
@@ -54,17 +54,19 @@
   },
   "homepage": "https://github.com/solidos/solid-ui",
   "dependencies": {
+    "@inrupt/universal-fetch": "^1.0.1",
```
SharonStrats, line 57: will leave it in dependencies instead of dev because I believe we will need it in other places so that we can use node 18.

SharonStrats: Thought I would leave it until we make a decision.

Note: this is a node 16 problem with the clearImmediate function affecting the autocompleteField tests.

package.json
```diff
@@ -105,7 +107,7 @@
     "html-webpack-plugin": "^5.5.0",
     "husky": "^7.0.4",
     "jest": "^27.5.1",
-    "jsdom": "^19.0.0",
+    "jsdom": "^21.1.1",
```
SharonStrats, line 110: This is something I tried, the tests still work with the upgrade. But we might want to revert back so wanted to call it out.

bourgeoa:
closed with https://github.com/SolidOS/solid-ui/pull/543/commits/8924ae81610399e968faf5b83a2f5669db70c87c

PR #544, opened 2023-04-17
Chat proof

bourgeoa:
@SharonStrats 

I think I have something to work with if we can begin to add tests
first test in test/chat/keys.test.ts is failing with 
```
crypto.getRandomValues must be defined
```


src/chat/keys.ts
```diff
@@ -118,61 +114,68 @@ const keyAclBody = (keyDoc, me) => {
...
+/**
+ * set ACL
+ * @param keyDoc
+ * @param aclBody
+ */
 async function setAcl (keyDoc, aclBody) {
   // Some servers don't present a Link http response header
-  // if the container doesn't exist yet, so refetch the container
-  // now that it has been created:
+  // if the container doesn't exist yet, so refetch the resource
```
TallTed, line 129:
```suggestion
  // if the container doesn't exist yet, refetch the resource

```

src/utils/keyHelpers/acl.ts
```diff
@@ -0,0 +1,78 @@
...
+  // FIXME: check the Why value on this quad:
+  debug.log(store.statementsMatching(store.sym(keyDoc), store.sym('http://www.iana.org/assignments/link-relations/acl')))
+  const keyAclDoc = store.any(store.sym(keyDoc), store.sym('http://www.iana.org/assignments/link-relations/acl'))
+  if (!keyAclDoc) {
+    throw new Error('Key ACL doc not found!')
+  }
+
+  // delete READ only keyAclDoc. This is possible if the webId is an owner
+  try {
+    const response = await store.fetcher.webOperation('DELETE', keyAclDoc.value) // this may fail if webId is not an owner
```
SharonStrats, line 24: @bourgeoa I wanted to find out more about why we do this. It looks like we try to delete the acl, but then "PUT" it again with the read turtle below. Can you tell me more about it?

test/unit/chat/signature.test.ts
```diff
@@ -0,0 +1,21 @@
...
+const PRIV_KEY = 'a11bc5d2eee6cdb3b37f5473a712cad905ccfb13fb2ccdbf1be0a1ac4fdc7d2a'
+const PUB_KEY = '023a9da707bee1302f66083c9d95673ff969b41607a66f52686fa774d64ceb87'
+
+describe('signature', () => {
+  it('serializeMsg', () => {
+    const msg = { id: 'test' }
+    const res = serializeMsg(msg)
+    expect(res).toEqual(JSON.stringify(msg))
+  })
+  it.skip('sign and verifySignature', () => {
```
SharonStrats, line 13: @angelo-v I'm having trouble with this test.  I've tried several different things. I was wondering if you can tell just by looking what I'm doing wrong. Here is the error.
`[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".] {
  code: 'ERR_UNHANDLED_REJECTION'`

src/utils/keyHelpers/acl.ts
bourgeoa: This is because the ACL exists and is Read only.
We cannot Write the new key with such ACL

test/unit/README.md
```diff
@@ -1,10 +1,40 @@
...
+There is an extension that can be used to aide in debugging jest tests. To find out more about it you can look at [Jest Runner](https://marketplace.visualstudio.com/items?itemName=firsttris.vscode-jest-runner)
+
+### Mocking
+In solid-ui we do not currently follow a MVC pattern therefore there can be some difficulty in testing. The following are patterns to help with this.
+#### Store methods
+##### Load
+In SolidOS we use [rdflib.js](https://github.com/linkeddata/rdflib.js/) to work with LinkedData. The way this works is that you first load the document you need to work with into the store. Once the document is loaded you can then access the data by using additional methods on the store such as `any, each,...`. Since the data that gets returned will need to be mocked, `load` doesn't need to do anything. See below for what you need to put in the top of your file in order to mock the `load` method.  
+`import { store } from 'solid-logic'` at the top of your file.
+
+`store.fetcher.load = jest.fn().mockImplementation(() => {})`
```
TallTed, line 36:
~~~suggestion
In SolidOS, we use [`rdflib.js`](https://github.com/linkeddata/rdflib.js/) to work with LinkedData. First, you load the document you need to work with into the store. Once the document is loaded, you can access the data by using additional methods on the store such as `any`, `each`, etc. Since the data that gets returned will need to be mocked, `load` doesn't need to do anything. You need to put the following into your file to mock the `load` method:
```
import { store } from 'solid-logic'      /* at the top of your file */
...
store.fetcher.load = jest.fn().mockImplementation(() => {})
```
~~~

test/unit/README.md
```diff
@@ -1,10 +1,40 @@
...
 
-### Mock the fetcher
+##### Any
 
+##### Each
```
TallTed:
```suggestion
##### `Any`

##### `Each`
```

TallTed (changes requested):
Minor grammar, punctuation, etc.

test/unit/chat/signature.test.ts
bourgeoa: I do not have your issue but `TypeError: Expected input type is Uint8Array (got object)`
This seems ajest/jsdom issue with asolution proposed in https://github.com/paralleldrive/cuid2/issues/44#issuecomment-1531731695

SharonStrats: Thanks @bourgeoa I will check that out.

src/utils/keyHelpers/acl.ts
```diff
@@ -0,0 +1,74 @@
...
+    throw new Error('Key ACL doc not found!')
+  }
+
+  try {
+    await store.fetcher.webOperation('PUT', keyAclDoc.value, {
+      data: aclBody,
+      contentType: 'text/turtle'
+    })
+  } catch (err) {
+    if (err?.response?.status !== 404) { throw new Error(err) }
```
SharonStrats, line 28: @bourgeoa I'm wondering, if it errors with a 404, should we create the resource in that case?

And otherwise if it errors would it be because the user is not logged in? If so I will mock that data for testing.

test/unit/README.md
SharonStrats: @TallTed thank you for all the corrections.

test/unit/README.md
```diff
@@ -10,30 +10,35 @@ The following command will run all the tests.
 `npm run test`
 
 ### One file at a time
-There are a lot of tests in `solid-ui` so you will most likely want to run only the test you are working on at the moment. In order to do that you can use the following command.
+There are a lot of tests in `solid-ui`, so you will most likely want to run only the test you are working on at the moment. You can use the commands like the following to do that:
```
TallTed, line 13: Oops!
```suggestion
There are a lot of tests in `solid-ui`, so you will most likely want to run only the test you are working on at the moment. You can use commands like the following to do that:
```

src/utils/keyHelpers/acl.ts
bourgeoa: PUT should never error with 404 so we should remove the if
The reason for failing could be 401 or 403 (not authenticated or not authorized)

src/utils/keyHelpers/accessData.ts
```diff
@@ -0,0 +1,80 @@
+import * as debug from '../../debug'
+import { CERT } from '../../chat/signature'
+import { store } from 'solid-logic'
+import * as ns from '../../ns'
+import { NamedNode } from 'rdflib'
+
+export const getPodRoot = async (webId: NamedNode) => {
```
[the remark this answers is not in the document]
bourgeoa: I did not knew of this function.
Well I think : 
- keep for know the one we have which is dealing with more cases : multiple storages or no satorage in WebID, no storage
- make the one we have compaible : I think it returns a NamedNode
- create an issue in solid-logic to improve the getPodRoot()

PR #545, opened 2023-04-19
removed unused vars for linting

SharonStrats:
Just cleaned up the lint errors.


timea-solid (approved):
I double checked and this is fine.

PR #549, opened 2023-05-25
remove getPodRoot

bourgeoa:


SharonStrats (approved):
I was also having a preference problem with the SolidOS chat, but wasn't sure yet if it was something in my setup.

PR #550, opened 2023-05-29
Chat thread with proof

bourgeoa:
result of merging  chat main (including message signature) and chat-threads.

This is nicely working with 2 remaining isssues  related to send : 
- no area to add a comment (send area) on a thread opened in the past
- on same day you cannot add new comments. Send fails and you must reload the chat


src/utils/keyHelpers/accessData.ts
```diff
@@ -68,25 +42,7 @@ export async function getKeyIfExists (webId: NamedNode, keyUrl: string, keyType:
     const key = store.any(webId, ns.solid(keyType)) // store.sym(CERT + keyType))
     return key?.value // as NamedNode
   } catch (err) {
-    if (err.response.status === 404) {
```
SharonStrats, line 71: @bourgeoa I took out this code because it only creates the document but it doesn't create a key, I believe we take care of this a few methods above in the call stack. Let me know what you think. I can always put it back.

timea-solid:
If this is also containing the chat-thread branch, I guess we can close that one?

package.json
```diff
@@ -55,6 +55,7 @@
   },
   "homepage": "https://github.com/solidos/solid-ui",
   "dependencies": {
+    "16": "^0.0.2",
```
timea-solid, line 58: I think this is by mistake here?

timea-solid (approved):
Good work!

PR #554, opened 2023-12-08
Allow pim:storage to be in a settings doc

michielbdejong:
This way, the profile doc could be public but the pim:storage link doesn't need to be.

Draft code, didn't test yet.


michielbdejong:
CC @janschill @thhck

PR #563, opened 2024-10-23
Add social media platform icons in a subdirectory

timbl:


timbl:
The publish  fails Error: Error: Unable to publish solid-ui v2.4.33-cb021f4d to NPM.   Maybe best to merge and hand publish to NPM

timbl:
merge without auto publish

melvincarvalho:
Sorry, bit late to the party here, but perhaps consider adding nostr svg logo too?  

Nostr is a social net used used by about 50k users, including Jack Dorsey and Ed Snowden.  

It would require 64 byte hex public key to link to (which could also be a urn/uri), and it's possible link back to solid using [NIP-05](https://github.com/nostr-protocol/nips/blob/master/05.md)

Purple on white:

https://github.com/mbarulli/nostr-logo/blob/main/SVG/nostr-icon-purple-on-white.svg

White on purple:

https://github.com/mbarulli/nostr-logo/blob/main/SVG/nostr-icon-white-on-purple.svg

PR #742, opened 2026-04-16
show buttons for taking photo etc

SharonStrats:
When trying to use for the new design of the profile pane I wasn't able to take a picture. I was only able to pause the video, there was a black button underneath, but that disappeared. I'm not sure how that was supposed to work exactly. Take Photo, Retake photo, and Use Photo are now visible and it works. I have only tested this when I copied the code over to profile pane and worked on it there. I've copied the changes exactly. Will test further.

### DIFF
--- a/src/media/media-capture.ts
+++ b/src/media/media-capture.ts
@@ -40,6 +40,10 @@ export function cameraCaptureControl (
   const div = dom.createElement('div')
   let destination, imageBlob, player, canvas
 
+  const setButtonVisible = (button: HTMLElement, visible: boolean) => {
+    button.style.display = visible ? 'inline-flex' : 'none'
+  }
+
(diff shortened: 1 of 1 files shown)


timea-solid (approved):

timea-solid:
@SharonStrats it looks good in general. Can you take a look into the Copilot review comments? I am not sure if your goal was to override the tooltip or not. I have a feeling Copilot is right, in this case in which this needs more work. 
I pre approved it to not block your work.

SharonStrats:
Yeah no problem. I need to review I made other changes to it when I copied it into profile pane I'll take a closer look and update and then use in profile pane as apart of post-milestone3m.

SharonStrats:
I created a web component from this logic to use in profile-pane for this so I'm thinking we just leave as is. We can always reopen if necessary.

PR #743, opened 2026-04-22
Refactor moving code to ui-forms and solid-ui-core

timbl:
The forms area part of solid-ui but should really be a separate product.  This refactor (done with the help of claude code) moves all the code around but changed nothing in the API of solid-ui.   The tests all run still as solid-ui tests, and all pass.


NoelDeMartin:
I've taken a look at this PR and the related issues, looking at the code it seems like these are very tied together to SolidOS right now (because they use things like `solid-logic`).

I suppose the reusable web components from https://github.com/SolidOS/solid-web-components shouldn't depend on that, though maybe they can depend on rdflib (this is related with the ongoing discussion to standardize Solid Web Components across different projects like PodOS, etc.). I guess these RDF forms should be reusable as well, not tied together to SolidOS.

About releasing them as a separate package or not, I don't have many opinions... Though in general, I'd prefer to keep as few repositories as possible because it's already very difficult to work across different repositories. If we were using a monorepo (hopefully some day we will!) it would be easier to separate things in individual packages, but until that happens I think it's better to rely on tree-shaking for things that share the same concern.

In any case, I think the biggest doubt I have at this point is how to handle global state. We can discuss this as we start building the solid-ui "design system" components, because we were saying they should be reusable as well (just not as stylable). But if they depend on global state (like `solid-logic`), they won't be reusable either.

And finally, I see these use the imperative DOM manipulation primitives, so at some point they should be migrated to Web Components as well (if we end up deciding on that).

timea-solid:
We decided to make Web Components out of RDF forms and keep them in solid-ui together with all web components we have.

PR #747, opened 2026-05-11
a11y: demote tab-widget body from <main> to <div>

m5x5:
## Summary
- The tab widget previously created `<main>` elements for both its outer body container and per-tab body. Combined with the shell's own `<main>`, each tabbed pane view had two or more `<main>` landmarks per page.
- The page-level `<main>` landmark now lives on the shell's `.app-view` (see SolidOS/mashlib#394). Tab widget internals become plain `<div>`s so we have exactly one `<main>` per page.
- No styling change: the widget uses `style.tabsMainElement` and `bodyMainStyle` inline, not a `main` selector.

Addresses SolidOS/profile-pane#310.

## Coordinated PRs
- SolidOS/mashlib#394 — Promote `.app-view` to the page-level `<main>` landmark (land first)
- SolidOS/profile-pane#312 — Drop inner `<main>` from ProfileView and EditProfileView

### DIFF
--- a/src/tabs.ts
+++ b/src/tabs.ts
@@ -187,7 +187,7 @@ export function tabWidget (options: TabWidgetOptions) {
   const navElement = rootElement.appendChild(dom.createElement('nav'))
   navElement.setAttribute('style', style.tabsNavElement)
 
-  const mainElement = rootElement.appendChild(dom.createElement('main'))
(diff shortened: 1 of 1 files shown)


timea-solid:
Tested locally and works. thanks

PR #748, opened 2026-05-14
[WIP] Implement design system button

NoelDeMartin:
This PR introduces a first draft for how we could implement components moving forward, following [the design system from Figma](https://www.figma.com/design/88xQawSSXzfQprEZPaKplZ/Solid-OS?node-id=1643-9433&t=te10zEWzYsMpAp8F-4).

Some things to note:

- Variable names are aligned with Figma naming, and included in this package (in `src/design-system/styles/main.css`). The current global variables should be removed from mashlib, and if needed for reuse, we could export them from this package in some way.
- I just placed them in a `design-system` folder temporarily, once we settle on the final conventions and start migrating all components we should think about folder organization and how to import/export various bundles.
- After some research into other projects using Lit ([Web Awesome](https://github.com/shoelace-style/webawesome/), [Home Assistant](https://github.com/home-assistant/frontend/)), it seems like it's standard practice to couple component registration with implementations (meaning, we will export bundles with side-effects). We can still think about this, but seeing how other projects do it, maybe we should use a different approach for tree-shaking that isn't removing side-effects.
- I have also implemented a proposal for using icons moving forward, see `src/design-system/lib/icons` and how icons are rendered (using `<icon-lucide-plus>`, etc.). The main drawbacks is that we still have to import individual icons, rather than being resolved individually. Otherwise we could import the entire icon collection. But I haven't found a way to resolve this dynamically using web components.
- I have created a new storybook for the new component, which is aligned with the examples in the Figma design. Looking over the other existing storybook entries, many seem broken or outdated; so maybe it would be a good idea to clean them all up and start from scratch. Ideally, every component variant/property combination would be showcased in storybook. Because of that, the current button implementation is quite simple (just has 3 variants and icon slots). I'd rather start like this and adding component complexity as needed, rather than trying to implement a complex button from the beginning.

So, this is still a draft but I think it's very close to what I'd like to have eventually. So now it would be a good point to give feedback if there is anything at all you don't like about this PR :).

PS: You can see a temporary deployment of the storybook for this code here: https://noeldemartin.github.io/solid-ui/?path=/docs/design-system-button--docs


timea-solid:
In general it looks fine for me. Thank you for doing the research.'
The icon usage seems a bit 'clunky' so I hope for a more elegant solution. 

Are we going to use https://github.com/SolidOS/solid-assets for global variables?

NoelDeMartin:
> The icon usage seems a bit 'clunky' so I hope for a more elegant solution.

Can you elaborate what you mean? Is it the fact that we have to import each icon individually? Or you don't like the idea of having a web component for each icon?

To clarify, this is how we'd use icons with this new approach:

```html
<solid-ui-button>
    Confirm
    <icons-lucide-check></icons-lucide-check>
</solid-ui-button>
```

If you're looking at the code in the storybook, that is indeed more confusing than this. But that's only because we are trying to render icons dynamically. And maybe it could be done more elegantly, but given that it's just storybook code I don't care about it as much as production code.

The real problem I would agree with is the need to import each icon, as I mentioned... But I'm not sure it's easy to solve with our current setup, unfortunately :(. Other than that, this is very close to [how I've been using icons in my apps for years](https://github.com/NoelDeMartin/umai/blob/a0d2887e8f70d93ee6f3b3d33b39452cea5874c9/src/components/modals/ShareRecipeModal.vue#L28) and the DX is pretty good (we could also configure that `<i-*` prefix instead of `<icons-*` if we want).

> Are we going to use https://github.com/SolidOS/solid-assets for global variables?

I didn't know the variables came from there, I thought they were [defined in mashlib](https://github.com/SolidOS/mashlib/tree/post-milestone3m/src/styles/themes). In any case, they should definitely be declared in a single place, and imported from whatever packages publishes them elsewhere. I would vote for exporting them from this package (`solid-ui`), given that it's where they're going to be used the most since this repository should define the foundations of the design system. But if we still want to keep it on solid-assets, that's fine, I can also make a proposal to show how we could export/import them to keep a single source of truth.

timea-solid:
Oh, ok. I took the storybook as example.
I like 
```
<solid-ui-button>
    Confirm
    <icons-lucide-check></icons-lucide-check>
</solid-ui-button>
```
And I think I am also fine with importing it separately. 

Regarding the https://github.com/SolidOS/solid-assets , we are not using it now. 
I was hoping this repo can be adopted by Solid orga itself, but I am not sure about it. And provide the CSS variables, raw for anyone who needs them. 
Do you need to know this in advance for our plan with solid-ui?

NoelDeMartin:
> Do you need to know this in advance for our plan with solid-ui?

Not really, I think we can proceed either way. But I'm not sure how useful it would be to have only the variables in that separate repo. If people want to reuse "Solid OS"-flavor UI, be it using CSS variables, web components, or anything else it would make sense that everything is exported from the same package, given that all of these are very coupled together.

As I said elsewhere, though, I wouldn't care so much about using different packages if things were in the same repository. But since we aren't using a monorepo, it complicates things a lot using different repositories. For example, if we start adding some automated tests that rely on any of the things inside a different repository, they'd constantly break in PRs until they are merged. Ideally, nothing with a broken CI should be merged, but if we have external dependencies we'll be forced to merge "broken" PRs and hope for the best, or be very disciplined about releasing related versions before merging, etc. All in all, I think it's not a good idea to have strong dependencies like this between repositories.

timea-solid:
You make perfect sense. Especially after thinking about it some more, let's proceed with solid-ui. I will keep the other repo, but we will probably delete it. 
Given that the icons are Lucide Icons and are available online, the more reason not to have another repo. In the past, the icons were imported from solid-ui through GitHub Pages. When GitHub was down or when we moved organizations, nothing worked anymore. This was the initial thought of making them separate. But now we have an entirely new solution.

NoelDeMartin:
I just pushed a couple of updates to this draft, here's a summary of the most relevant:

- Added the tailwindcss preflight. This only uses the preflight utilities to make it easier to style our components, we could also use another reset like [modern-normalize](github.com/sindresorhus/modern-normalize), etc. It doesn't mean we're going to use tailwind to style anything.
- I have reimplemented the Icons Web Component compiler. I did this because trying to use icons in different components they weren't easy to style or define sizes. Still a draft and I'm open to suggestions about this.
- There are 2 base css files now, `variables.css` and `common.styles.css`. The first one should be imported alongside the library in the main page of an app using this library, and it contains global styles (for now just variables). The second file should be imported to each component, and contains resets and reusable utilities inside our components (there isn't any utility yet). This will take advantage of [Constructable Stylesheets](https://web.dev/articles/constructable-stylesheets), so the styles will be reused even if they are scopped to each component.
- I have opened a follow-up PR #761 with "primitive" components. Basically, those are the ones intended to be used in any Solid project, not just Solid OS or with the Solid UI styles.

timea-solid:
I was thinking: maybe it would be valuable to finalize the button example and use it in profile-pane already. Meaning: that we ship a first SolidOS component, accessible and styled and all. 
What do you think?

NoelDeMartin:
> maybe it would be valuable to finalize the button example and use it in profile-pane already.

Yes, that'd be ideal, but I was waiting until we start using the new branching strategy to do it. Two weeks ago we said it would take about a week, so I guess we should try to get it done this week already 😅. My main blocker right now is the #744 PR, until that one is merged (in `staging` or wherever, doesn't have to be released), I'm trying not to do anything "final". But once that is merged in, I can open a real PR that can be merged and also replaces the old button wherever possible. There'll probably be some time until we stop using it everywhere, but that's ok.

NoelDeMartin:
Superseeded by #775

PR #760, opened 2026-05-20
replace Inrupt by Uvdsl OIDC client

### LINKED ISSUE #274: Switch out authentication library
We will be using https://github.com/uvdsl/solid-oidc-client-browser

Initial ticket over at: https://github.com/solidos/solidos/issues/284

See a code example in LDO: https://github.com/o-development/ldo/commit/2aab9b3f31b2e0b1152787b2c9e8b45760dfb8b2

bourgeoa:
# Commit Notes - solid-ui (uvdsl contract adoption)

## Scope
Update solid-ui to consume the new uvdsl session contract directly and stabilize local integrated development with linked solid-logic.

## What changed
- Migrate login calls to new signature:
  - from object-style `{ oidcIssuer, redirectUrl }`
  - to positional `login(issuer, redirectUrl)`
- Replace auth session EventEmitter usage with EventTarget usage (`addEventListener('sessionStateChange', ...)`) where updated.
- Switch session state reads to the new contract (`webId` / `isActive`) with fallback only where still needed for integrated migration.

## Local integrated development notes kept intentionally
`tsconfig.json` includes settings to avoid duplicate rdflib type identities when solid-ui is linked with local solid-logic:
- `preserveSymlinks: true`
- `baseUrl` + `paths` mapping for `rdflib`

This avoids false-positive type errors like `LiveStore is not assignable to IndexedFormula` caused by two rdflib type origins.

## Jest stabilization for local linked repos
- Map `solid-logic` to source in Jest to avoid loading webpack bundle output during tests.
- Mock `@uvdsl/solid-oidc-client-browser` for Jest/JSDOM to avoid ESM/import.meta runtime issues.

## Integration outcome
- solid-ui now listens to the native session state change flow expected by the uvdsl client.
- Downstream packages can continue to function while solid-logic carries temporary compatibility shims.

### Title
feat(auth): adopt uvdsl session contract in solid-ui and stabilize local linked dev

### Body
- Update solid-ui auth flow to uvdsl login/session contract
- Move session listeners to EventTarget sessionStateChange pattern
- Keep local linked-repo TypeScript settings to prevent duplicate rdflib type identity errors
- Add Jest mappings/mocks so tests run against source in local integrated workspace
- Document local integrated development rationale in README

### Title
refactor(auth): remove remaining transitional auth fallbacks in solid-ui

### Body
- Remove temporary fallback reads for legacy session shapes
- Keep only uvdsl-native state/event handling once integrated tests are stable

### DIFF
--- a/src/design-system/lib/auth/SolidAuth.ts
+++ b/src/design-system/lib/auth/SolidAuth.ts
@@ -13,11 +13,13 @@ export default class SolidAuth implements AuthContext {
   constructor (public signupUrl: string = DEFAULT_SIGNUP_URL) {}
 
   get account (): Account | null {
-    if (!authSession.info?.isLoggedIn || !authSession.info?.webId) {
+    const sessionAny = authSession as any
+    const webId: string | undefined = sessionAny.webId ?? sessionAny.info?.webId
+    const isActive: boolean = sessionAny.isActive ?? sessionAny.info?.isLoggedIn ?? Boolean(webId)
+    if (!isActive || !webId) {
       return null
     }
 
-    const webId = authSession.info.webId
     const me = solidLogicSingleton.store.sym(webId)
     const avatarUrl = findImage(me) ?? undefined
 
@@ -44,10 +46,7 @@ export default class SolidAuth implements AuthContext {
 
     locationUrl.hash = ''
 
-    await authSession.login({
-      redirectUrl: locationUrl.href,
-      oidcIssuer: loginUrl
-    })
+    await (authSession as any).login(loginUrl, locationUrl.href)
   }
 
   async signup () {
@@ -59,14 +58,27 @@ export default class SolidAuth implements AuthContext {
   }
 
   onSessionUpdated (callback: () => unknown) {
-    authSession.events.on('login', callback)
-    authSession.events.on('logout', callback)
-    authSession.events.on('sessionRestore', callback)
+    const sessionEventTarget = authSession as unknown as EventTarget
+    const sessionAny = authSession as any
+    const listener = () => {
+      callback()
+    }
+    if (typeof sessionEventTarget.addEventListener === 'function') {
+      sessionEventTarget.addEventListener('sessionStateChange', listener)
+    } else {
+      sessionAny.events.on('login', callback)
+      sessionAny.events.on('logout', callback)
+      sessionAny.events.on('sessionRestore', callback)
+    }
 
     return () => {
-      authSession.events.off('login', callback)
-      authSession.events.off('logout', callback)
-      authSession.events.off('sessionRestore', callback)
+      if (typeof sessionEventTarget.removeEventListener === 'function') {
+        sessionEventTarget.removeEventListener('sessionStateChange', listener)
+      } else {
+        sessionAny.events.off('login', callback)
+        sessionAny.events.off('logout', callback)
+        sessionAny.events.off('sessionRestore', callback)
+      }
     }
   }
 }
--- a/src/design-system/lib/auth/SolidAuth.ts
+++ b/src/design-system/lib/auth/SolidAuth.ts
@@ -27,9 +27,8 @@ export default class SolidAuth implements AuthContext {
   constructor (public signupUrl: string = DEFAULT_SIGNUP_URL) {}
 
   get account (): Account | null {
-    const sessionAny = authSession as any
-    const webId: string | undefined = sessionAny.webId ?? sessionAny.info?.webId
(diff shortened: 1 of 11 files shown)


NoelDeMartin (changes requested):
Looking at the conflicts, the only one I'm aware of is the Header, given that I changed it in my recent work with the design system. It should be fairly easy to resolve the conflicts, though, because I just added some conditionals but the Header component itself hasn't changed a lot.

The only important remark is that you should also add your modifications to the Auth workflow in `src/design-system/lib/auth/SolidAuth.ts`. It's not ideal that we have it in two places, but once we remove the `DESIGN_SYSTEM_HEADER_ACCOUNT` feature flag, it'll be just one (See #771 for details).

Finally, the history of the PR is a bit confusing to review because it includes even some commits from `main`, and I'm not sure what else. I'd suggest rebasing or cherry-picking your changes on top of the `staging` branch to make it clearer. That could also resolve most of the conflicts.

bourgeoa:
@NoelDeMartin I introduced your 3 comments

NoelDeMartin:
@bourgeoa I don't see any changes in `src/design-system/lib/auth/SolidAuth.ts`, did you look into that?

bourgeoa:
@NoelDeMartin done. I missed it somehow

src/design-system/lib/auth/SolidAuth.ts
```diff
@@ -44,10 +59,7 @@ export default class SolidAuth implements AuthContext {
 
     locationUrl.hash = ''
 
-    await authSession.login({
-      redirectUrl: locationUrl.href,
-      oidcIssuer: loginUrl
-    })
+    await (authSession as any).login(loginUrl, locationUrl.href)
```
NoelDeMartin: We should stop using `any` unless there is a very good reason for it (such as using a 3rd party library that is poorly typed, or some encapsulated type check that is not worth typing). In this case, since this comes from `solid-logic` which is a library we control, it should be possible to type correctly, right? It's also something quite important, so I would avoid `any` to reduce the potential for bugs as much as possible.

src/v2/components/layout/header/Header.ts
```diff
@@ -14,6 +14,36 @@ const DEFAULT_SOLID_ICON_URL = 'https://solidproject.org/assets/img/solid-emblem
 const DEFAULT_SIGNUP_URL = 'https://solidproject.org/get_a_pod'
 const DEFAULT_LOGGEDIN_MENU_BUTTON_AVATAR = icons.iconBase + 'emptyProfileAvatar.png'
 
+async function clearPersistedAuthState (): Promise<void> {
```
NoelDeMartin, line 17: There is some functionality, such as this function, that is not included in `SolidAuth`. Any reason why?

tsconfig.json
```diff
@@ -64,7 +64,9 @@
     ] /* List of folders to include type definitions from. */,
     // "allowSyntheticDefaultImports": true,  /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
     "esModuleInterop": true, /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
-    // "preserveSymlinks": true,              /* Do not resolve the real path of symlinks. */
+    "preserveSymlinks": true, /* Do not resolve the real path of symlinks. Needed for local linked solid-logic. */
+    "baseUrl": ".", /* Base directory to resolve non-absolute module names. Needed for paths mapping. */
+    "paths": { "rdflib": ["./node_modules/rdflib"] }, /* Map rdflib to avoid duplicate type identity when linked with solid-logic. */
```
NoelDeMartin, line 69: Why is this necessary? Doesn't the default resolution already resolve `rdflib` to the node_modules path?

NoelDeMartin (changes requested):

tsconfig.json
bourgeoa: This is because in local development with solid-logic I ended with conflict versions of rdflib

src/v2/components/layout/header/Header.ts
bourgeoa: The night was bad. Done

src/design-system/lib/auth/SolidAuth.ts
bourgeoa: Resolved in solid-logic https://github.com/SolidOS/solid-logic/pull/292

NoelDeMartin (approved):
I'll probably change some of the changes here in my PR #780, like the rdflib alias which I'm not sure is entirely necessary yet. But we can merge as is for now and clean it up later.

PR #761, opened 2026-05-21
[WIP] Implement reusable primitives

NoelDeMartin:
This is a follow up from #748, and it implements some reusable "primitive" components alongside a new Solid UI component (`<solid-ui-account>`).

Here's a summary of the most relevant points:

- The first commit can be ignored (assuming you've looked at #748)
- You'll notice the "primitive" components are very narrow in behaviour (login, signup, and logout buttons are separate components, etc.). Initially I started with more useful components, but I realized it's very difficult to make them useful whilst still being "stylable". Because of that, I ended up doing smaller components and the more complete components sit on the "design system" layer. We can talk about that, but I think it's the best approach for now. The markup for primitive components can be overriden completely, look at their storybooks for examples.
- One of the key architecture decisions here has been to decouple the Authentication from components. In fact, you'll notice the only authentication that has been wired up is Storybook's (which is a stub). We discussed this in [this week's meeting](https://solidos.solidcommunity.net/public/SolidOS%20team%20meetings/2026/2026_05_20.md), but TLDR, if we really want these components to be reusable, we have to decouple them from the actual authentication (Solid UI uses `solid-logic`, but other applications won't).
- I have been struggling to implement a decent Popover (and have also been experimenting with Dialogs) with Web Components. This is one of the challenges I mentioned initially about web components. In particular, the targeting and relations between different elements inside and outside of the Shadow DOM are difficult to manage (a button inside a component is supposed to open a menu from a different component, possibly even from the light DOM, etc.). We'll need to keep investigating this, but I wonder if it would make sense to rely on third party libraries for these type of interactions (maybe [Web Awesome](https://webawesome.com/), given that it's very aligned with our tooling?).
- Related to the previous point, I'm aware that some of the components (specially the `<solid-ui-menu>`) are not very accessible. This is something I want to continue exploring as we work on these.
- I am also aware that I'm using some advanced features that may not be supported in older browsers, such as [the `:state()` selector](https://caniuse.com/wf-state) or [the Popover API](https://caniuse.com/wf-popover). I have already thought on some ways to polyfill them, or even replace them, but for now I think it's ok to use them until we decide on the final direction.
- Check out [the updated Storybook](https://noeldemartin.github.io/solid-ui/) to see al the components, both design system and primitives.


timea-solid:
Interesting.
I suppose the design is not final on them. You made a component Account? I would have not thought of that. I guess I can add more elements to the menu when logged in.
And I do not get what Guard is.

NoelDeMartin:
> I suppose the design is not final on them.

I did try to follow the design from Figma, but maybe some things can still be improved (specially the drop down). I also tried to make it reusing the button styles that we have. For example, I am aware the "sign up" button is different than the design, but I used the secondary variant that is defined in the "design system" section.

Yes, we can always make buttons or any UI that is outside of the design system, but there should be a reason for it. Maybe we can discuss this with the designers now that we've opened a discussion about the design system.

> You made a component Account? I would have not thought of that. I guess I can add more elements to the menu when logged in.

Yes, this component is SolidOS-specifc, or "solid-ui", so we'll add more things in there. However, we still want this to be reusable so we'll have to think of a way to do that customizable in some way. I added a storybook for the Menu component, so it could be customized in a similar way (using an "options" slot for example).

> I do not get what Guard is.

This component may not be necessary, but I thought it would be useful. Initially, I wanted to create the "Account" component directly as a primitive, but then I realized it had too many opinions. So I split it into smaller components that can be combined to create it on the Solid UI layer.

The "guard" component in particular is just a flow control component to show different content for logged in users or guests. In the end I didn't use it because Account needs to access the `auth` anyways, but I still think it could be useful in some situations. Basically, it's a way to enable people to write an if switch in the UI declaratively, without needing to write JavaScript to access the authentication state, etc.

I guess it's a bit YAGNI, but it's such a small component and it can be useful so I think it should be fine to keep it. The name can be changed, of course, I wasn't sure about it either.

timea-solid:
@NoelDeMartin can you address with the designer the usage of too many different button styles?

NoelDeMartin:
Superseeded by #775

Issue #774, opened 2026-05-29
Export new Web Components


NoelDeMartin:
As part of #769, we've created a bunch of new components under "design-system" and "primitives". However, these are only used internally from now. The idea was that they could be reused in other projects as well, but we still have to decide how we want to export them, and how we migrate from the previous set of components (for example, there was also a "button" component before).

In particular, the "primitives" components are supposed to be usable in any Solid project, even without using `solid-ui`, so they're probably going to be released in a different package. But we still haven't decided which one.

NoelDeMartin:
I forgot we had this issue open 😅 Changed this to "ready" since this has already been merged.

In case someone is just seeing this issue, we decided to release everything in a single `solid-ui` package instead of making a distinction. See #790.

PR #775, opened 2026-05-29
#769 Update account components in header

### LINKED ISSUE #769: Replace Header with Design System components
As part of the ongoing work to define a Design System, we'll start by rebuilding the Header component.

NoelDeMartin:
This PR introduces the new web components we've been discussing over the last few weeks.

Here's some things to note:

- Related changes in mashlib are only limited to including the CSS for now (See PR https://github.com/SolidOS/mashlib/pull/409).
- The new functionality can be disabled with a feature flag. I'm doing this because even though the code is correct and mergable, it's not complete. There are still a couple of UI improvements I'm aware of, and I've opened a couple of related issues that should be tackled before we consider this feature "done". In case we want to make a release before this work has been completed, we could just deactivate the feature flag.
- I did include one commit that was completely AI generated, check out the commit message and formatting to see if we want to do that moving forward.
- This is all still internal for now, eventually we'll have to decide how to export these new components and possibly extract the "primitives" into a new package. The "design system" and "primitives" nomenclature is just a placeholder for now (See #774).
- Check out the issues I've opened for the remaining work (besides UI improvements):
    - https://github.com/SolidOS/solid-ui/issues/771
    - https://github.com/SolidOS/solid-ui/issues/772
    - https://github.com/SolidOS/solid-ui/issues/773
    - https://github.com/SolidOS/solid-ui/issues/777
- Check out the previous draft PRs where this work was already discussed:
    - https://github.com/SolidOS/solid-ui/pull/748
    - https://github.com/SolidOS/solid-ui/pull/761


timea-solid:
I would expect each component to have a Readme as well, for showcasing how to use it. (not just in Storybook). I fail to see how I would change the default sign up link.
Also, should each component have a test?

I see you removed the storybook stories we had. I would still need the forms stuff at least in - and once I move out the rdf forms part of the code, than delete it.

NoelDeMartin:
I haven't thought much about documentation yet, but the Storybooks already generate a Docs section with the current config (I updated my fork's deployment again: https://noeldemartin.github.io/solid-ui). Maybe that could be a follow-up task?

Regarding the storybooks I deleted, I thought all the components will need to be migrated to the new conventions so I preferred to clean up the files so that the storybook is up to date (and avoid confusing designers or anyone else who looks at the storybook). Keep in mind that "deleting" something in git is not really deleting it, we always have the history and can always restore it or see what it contained. Do you still think we should keep some of the old stories around?

timea-solid:
Very nice and clean commit history. I also like the AI related commit. Nicely done. I will use it as an example in our guidelines.

NoelDeMartin:
> I fail to see how I would change the default sign up link.

By the way, this can be configured in the `<solid-ui-provider>` component, which is not in the storybook because it's just a logic component (it doesn't have any UI). Instead of taking the signupUrl directly, it should probably take a config option, though. Or maybe we want to set the configuration using a global object in `window` or something. That's something we should decide eventually, and document it, but for now I think it's ok to have it there as we keep building the design system.

> Also, should each component have a test?

Well, not every component *should* have a test, but certainly each component *can* have a test. Ideally, I like to avoid redundant tests, for example I'm not much of a fan of writing tests like this one we have for the Header:

```ts
  it('is defined as a custom element', () => {
    const defined = customElements.get('solid-ui-header')
    expect(defined).toBe(Header)
  })
  ```
  
I prefer to test most UI using E2E tests that actually use the application for real. Still, I do agree with having some component unit tests for testing edge cases and so on.

Looking at the components I introduce in this PR, I don't think we need any specific test. Is there any that you had in mind?

I do agree that we're lacking the E2E tests, though. I have created an issue to look into this #777.

timea-solid:
Removing the storybook stories is fine.

timea-solid (approved):

PR #779, opened 2026-06-01
Integrating Tim's work on separating RDF forms

timea-solid:


timea-solid:
We decided to go against this approach and create web components out of RDF forms in solid-ui.

PR #787, opened 2026-06-04
#780 Migrate to Vite, Vitest, and Storybook 10

### LINKED ISSUE #780: Improve build system
The current build system has a series of issues that should be resolved, in particular:

- Component bundles are rolled-up, meaning that component used by more than one wrapper component throw an error on import.
- The build config is spread throughout multiple build tools depending on the environment: Babel, Webpack, Jest, Storybook, etc.
- Components have to be declared individually in `package.json` (which is automated with a script, but still not ideal).

Ideally, we'll solve this by migrating to [Vite](https://vite.dev/); which should also serve to modernize the tooling in the repository. However, if it isn't possible (or seems like too much work), we should at least fix these with the current tooling.

NoelDeMartin:
This PR migrates the build system from webpack to Vite, along with some related tooling (Vitest, Storybook, etc.).

Other than migrating the build config, it also changes how files are compiled to better align with best practices. You can find more details about this in the commit messages and inline comments.

There were also a couple of improvements I'm aware of that I decided to leave for the future:
- https://github.com/SolidOS/solid-ui/issues/788
- https://github.com/SolidOS/solidos/issues/300

Related PRs:
- https://github.com/SolidOS/mashlib/pull/419
- https://github.com/SolidOS/profile-pane/pull/398


SharonStrats:
I'm looking forward to using it when running watch pivot and see ing if I run into less issues.

package.json
```diff
@@ -2,135 +2,51 @@
...
 	"files": [
 		"dist/",
 		"README.md",
-		"LICENSE",
-		"src/design-system/styles/variables.css",
-		"src/primitives/styles/variables.css"
+		"LICENSE"
 	],
 	"scripts": {
-		"clean": "rm -rf ./dist ./src/versionInfo.ts ./docs/api .tsbuildinfo",
```
timea-solid, line 109: Is versionInfo replaced by tsbuildinfo?

package.json
```diff
@@ -2,135 +2,51 @@
...
-		"build-version": "sh ./timestamp.sh > src/versionInfo.ts && eslint 'src/versionInfo.ts' --fix",
-		"prebuild-js": "rm -f .tsbuildinfo",
-		"build-js": "tsc",
-		"sync-component-exports": "node scripts/sync-component-exports.mjs",
-		"postbuild-js": "rm -f dist/versionInfo.d.ts dist/versionInfo.d.ts.map && node scripts/build-component-dts.mjs",
-		"build-dist": "webpack --progress",
-		"build-form-examples": "npm run build-js && npm run build-version && npm run build-dist && cp ./dist/solid-ui.js ./docs/form-examples/",
+		"clean": "rm -rf ./dist ./docs/api .tsbuildinfo",
+		"build": "npm run clean && npm run build:npm && npm run build:styles && npm run build:cdn && npm run build:cdn-legacy",
+		"build:cdn": "vite build --mode cdn",
```
timea-solid: I introduced a convention in all repo that the scripts are all called with '-' instead of ':'. If it is a preference option here, I would keep it '-' to be in line with all CIs.

package.json
```diff
@@ -2,135 +2,51 @@
...
+		"build:cdn-legacy": "vite build --mode cdn-legacy",
+		"build:npm": "vite build",
+		"build:styles": "vite build --mode styles",
```
timea-solid: Can you please explain why the styles are separate?

package.json
```diff
@@ -2,135 +2,51 @@
...
-		"build": "npm run clean && npm run sync-component-exports && npm run typecheck && npm run build-version && npm run build-dist && npm run build-js && npm run postbuild-js && npm run build-storybook",
-		"build-version": "sh ./timestamp.sh > src/versionInfo.ts && eslint 'src/versionInfo.ts' --fix",
-		"prebuild-js": "rm -f .tsbuildinfo",
-		"build-js": "tsc",
-		"sync-component-exports": "node scripts/sync-component-exports.mjs",
-		"postbuild-js": "rm -f dist/versionInfo.d.ts dist/versionInfo.d.ts.map && node scripts/build-component-dts.mjs",
-		"build-dist": "webpack --progress",
-		"build-form-examples": "npm run build-js && npm run build-version && npm run build-dist && cp ./dist/solid-ui.js ./docs/form-examples/",
+		"clean": "rm -rf ./dist ./docs/api .tsbuildinfo",
+		"build": "npm run clean && npm run build:npm && npm run build:styles && npm run build:cdn && npm run build:cdn-legacy",
```
timea-solid: Also as a convention across all CI of all repos, we do the typecheck inside the build. Is there a reason to separate?

package.json
```diff
@@ -2,135 +2,51 @@
...
-		"watch:component-exports": "node scripts/watch-component-exports.mjs",
-		"watch:dist": "webpack --watch --mode development",
-		"watch": "npm run sync-component-exports && npm run build-version && npm run build-js && npm run postbuild-js && sh -c 'npm run watch:js & npm run watch:component-dts & npm run watch:component-exports & npm run watch:dist & wait'",
-		"dev": "npm run sync-component-exports && npm run build-version && sh -c 'npm run watch:js & npm run watch:component-dts & npm run watch:component-exports & npm run watch:dist & wait'",
+		"test": "vitest run",
+		"test-coverage": "vitest run --coverage",
+		"test-debug": "vitest --inspect-brk --no-file-parallelism",
 		"doc": "typedoc --out ./docs/api/ ./src/ --excludeInternal",
 		"prepublishOnly": "npm run build && npm run lint && npm test && npm run doc",
-		"preversion": "npm run sync-component-exports && npm run lint && npm run typecheck && npm test",
```
timea-solid, line 133: You asked why we have these. Alain is using them for releasing and as a convention we have them in all CIs of all repos.

package.json
```diff
@@ -179,61 +95,55 @@
...
 		"globals": "^17.6.0",
 		"isomorphic-fetch": "^3.0.0",
-		"jest": "^30.4.2",
-		"jest-environment-jsdom": "^30.4.1",
 		"jsdom": "^28.1.0",
-		"lit-css-loader": "^4.0.1",
 		"neostandard": "^0.13.0",
 		"nock": "^15.0.0",
 		"patch-package": "^8.0.1",
+		"path-browserify": "^1.0.1",
```
timea-solid, line 136: Can you explain why this dep?

package.json
```diff
@@ -179,61 +95,55 @@
...
 		"@testing-library/user-event": "^13.5.0",
-		"@types/jest": "^30.0.0",
 		"@types/jsdom": "^28.0.3",
 		"@types/node": "^25.8.0",
 		"@typescript-eslint/parser": "^8.59.3",
-		"babel-jest": "^30.4.1",
-		"babel-loader": "^10.1.1",
-		"css-loader": "^7.1.4",
+		"@vitest/coverage-v8": "^4.0.18",
+		"concurrently": "^10.0.3",
```
timea-solid, line 123: Can you explain why this dep is needed?

vite-config/babel.ts
```diff
@@ -0,0 +1,47 @@
...
+            // See https://lit.dev/docs/components/decorators/#using-decorators-with-babel
+            {
+                include: [
+                    'src/design-system',
+                    'src/primitives',
+                    'src/storybook',
+                ],
+                assumptions: { setPublicClassFields: false },
+                plugins: [
+                    ['@babel/plugin-transform-typescript', { allowDeclareFields: true }],
```
timea-solid, line 39: Before, we were handling these all in typescript config.

src/lib/env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
timea-solid, line 1: was this comment (and empty file) intentional?

vite-config/cdn.ts
```diff
@@ -0,0 +1,125 @@
+import { copyFileSync } from 'node:fs'
```
timea-solid, line 1: Why are some configs in this folder and some not?

timea-solid:
I want to talk about the choice of the lib folder name. I am not sure the name is good because - in the past dist was lib and it might be confusing. Lib may be reserved for the build structure. I do not mean to change the dist to lib, but simply not to use it.

package.json
NoelDeMartin: No, I looked at [the commit that introduced versionInfo](https://github.com/SolidOS/solid-ui/commit/f0b98df74f3ada1a4481a9a9bb9ec7b7fb4d1bb0) to understand what it is, and it doesn't seem to be used anymore. When it was introduced, this file was imported in the index, but it doesn't seem like it's used anymore (not sure when it was removed). So I thought it was dead code. Isn't it?

package.json
NoelDeMartin: Yes, these styles are the global styles that should be imported when using the library. What is documented as the `styles.css` `<link>` tag in the README, or the `@import "solid-ui"` from CSS. These are only variables, though, and these are the ones that could be overriden by the consumer of the library to customize theming. This is analogous to [Web Awesome's Design Tokens](https://webawesome.com/docs/tokens#customizing-design-tokens).

In contrast to that, the `.styles.css` files used for components are bundled in the javascript with each component bundle.

package.json
NoelDeMartin: Yes, this is used to polyfill the `'path'` dependency used by `mime-types` in `src/widgets/dragAndDrop.js`.

Webpack was [turning off that polyfill](https://github.com/SolidOS/solid-ui/blob/c59efc04ca8e7a10417cc80d2a53638963739e70/webpack.config.mjs#L50C5-L50C30), which resulted in an empty module that probably failed at runtime. If that really isn't necessary or used, we should just remove the code in dragAndDrop. But otherwise, the previous CDN build wasn't working.

package.json
NoelDeMartin: Yes, to run the `watch` script that coordinates watching both the default build and the styles build for development. 

It could also be implemented without a dependency, doing something like `npm run watch:npm & npm run watch:styles`, but the output produced by that isn't as readable, and killing the process leaves the first watch running.

src/lib/env.d.ts
NoelDeMartin: Yes, this is a [Triple-Slash Directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html), it serves to give types for vite APIs (like `import.meta.glob` used in `src/components/index.ts`).

vite-config/cdn.ts
NoelDeMartin: For code organization. The only config that is not in this folder is the default build config, but everything else is split in different files for code reuse or better legibility. If you have some suggestion for improvements, I can change it. But yes, it could potentially all be inside of `vite.config.ts` if we wanted.

vite-config/babel.ts
NoelDeMartin: Yeah, this is one of the most confusing parts of the whole build system 😅. But here's how I understand it.

Basically, Vite uses [Oxc](https://oxc.rs/) to compile from TypeScript to JavaScript, and Babel is used on top of that to convert modern JavaScript into older versions. This second pass of babel produces different outputs depending on the mode: CDNs transpile it into a version of JS supported by the browsers we target, but the default config only uses Babel to process [Lit decorators](https://lit.dev/docs/components/decorators/#decorators-babel) (I decided to use standard decorators, instead of experimental, because that's what's more likely to be supported in browsers some day). The reason why the default build doesn't do such aggresive transpilation is that npm consumers are expected to run their own bundling pipeline with their own browser support (this is also documented in [Lit's docs](https://lit.dev/docs/tools/publishing/#don't-bundle-minify-or-optimize-modules)).

In any case, yes it is confusing. I've been trying to graple with this and ask LLMs to make sure that everything is ok, but I'm still not 100% sure this is the most optional set up 😅. If you want to try to improve it, we could change it. But I'm not sure this can be removed from the babel config, because the CDN build doesn't even read the tsconfig at all (that is only used in the default build to produce the declaration files).

package.json
NoelDeMartin: Ok, I changed it.

package.json
NoelDeMartin: No particular reason, just to keep things separate and speed up the build script. Anyways, I added it back to stick to the convention.

package.json
timea-solid: The idea was that we have a way to debug which exact versions are used. This is an old tooling in all repos, which I kept when restructuring the cleaning the code. NOw, I could be wrong about removing it form index.
As I understand: Many modern repos prefer generating JSON or injecting version metadata via bundler/define plugins instead of writing a source file back into.

So, because this PR is about the build process, I would keep this in and discuss at a meeting if we need such a debug level for all repos or maybe just for mashlib. And than make a dedicated PR.

package.json
timea-solid: Got it. So this is the theming part of the design system. Which we decided to keep in solid-ui (and not mashlib or a separate repo).

package.json
timea-solid: Yeah, right! Thank you for reminding me. We keep the code for now.

vite-config/babel.ts
timea-solid: ok... We should add this to the repos, readme, to have it captured better than in a ticket. Pls add.

src/lib/env.d.ts
timea-solid: Maybe add a comment in the file not to be deleted.

vite-config/cdn.ts
timea-solid: In this case I would rename the folder: vite-config. Because we also have typescript configs and others in the root.

timea-solid:
A remark: at this point, this PR breaks the build of all panes using solid-ui (that is pretty much all). As soon as you change the location of files (such as lib/) all panes that import stuff will not find them. I would not chnage any 'older' code in this drastical way. 
Pls keep this PR only for build changes more or less.

package.json
NoelDeMartin: Yeah, I think the way to go if we want this kind of functionality is to just have a build plugin that generates this, rather than adding that into the source code.

About this PR, though, I'm not really removing any feature... I'm just cleaning dead code :). Isn't that always a good thing to do? Anyways, I can add it back if you want, but it's not doing anything. If you don't want to forget about this, we could create an issue and add it to the roadmap/weekly meetings, but I'm always in favor of deleting dead code.

src/lib/env.d.ts
NoelDeMartin: Why do you think this deserves a comment? It's a standard typescript feature, no different than imports or using .d.ts files. Which features need explaining and which don't?

It should be safe from deleting, though, because if you delete it you'll get type errors in CI (the `import.meta.glob` as I mentioned).

vite-config/babel.ts
NoelDeMartin: What do you mean "better than a ticket"? A ticket for what? And I'm not sure what you want me to add to the README, some documentation on how the build system works? I'm not sure it's a good idea to add this to the README, the issues with this type of documentation is that it becomes stale very quickly. Whenever we change something, we need to remember to also update the docs, etc. If anyone wants to learn about this, they should be able to trace it on the git history, and I have already added some pointers to that Lit documentation in the commit message.

package.json
timea-solid: I suggest to leave it in. This work spans all repos.

vite-config/babel.ts
timea-solid: It is better in a readme than in some comments on this ticket, which can get lost.
I prefer the readme for newcomers and all.

src/lib/env.d.ts
timea-solid: Ok. Leave as is.

package.json
NoelDeMartin: Ok, I added it back.

vite-config/babel.ts
NoelDeMartin: Ok, I added a new "Build Config" section to the README, let me know if it's clear.

timea-solid (approved):

timea-solid:
> A remark: at this point, this PR breaks the build of all panes using solid-ui (that is pretty much all). As soon as you change the location of files (such as lib/) all panes that import stuff will not find them. I would not chnage any 'older' code in this drastical way. Pls keep this PR only for build changes more or less.

I tested everything locally and it works. The lib folder also works. We can keep it.

bourgeoa:
I begun to look at vite implementation.

The main question is about the CDN.
I feel that that that changing to include solid-logic and/or rdflib must be on both ESM and legacy.
Anyway I think we are going to a major semver.

I will need some time to understand the way vite and vitest works. build/test performance is clearly there.

Not knowing how to commit dependencies updates/removed I created https://github.com/SolidOS/solid-ui/pull/794
There is only eslint that is not updated to 10. It should be soon when neostandard is upgraded

bourgeoa:
I tried to build profile-pane with solid-ui and got 37 errors of same type 
```
src/sections/bio/BioSection.ts:2:8 - error TS2882: Cannot find module or type declarations for side-effect import of 'solid-ui/components/actions/button'.

2 import 'solid-ui/components/actions/button'
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
The AI propose different solutions
```
The mismatch: profile-pane imports use nested category paths that don't match solid-ui's flat build output.

profile-pane imports	solid-ui exports resolves to	Actual dist
solid-ui/components/actions/button	dist/components/actions/button/index.d.ts	index.d.ts
solid-ui/components/forms/select	dist/components/forms/select/index.d.ts	index.d.ts
solid-ui/components/forms/combobox	dist/components/forms/combobox/index.d.ts	index.d.ts
solid-ui/components/media/photo-capture	dist/components/media/photo-capture/index.d.ts	index.d.ts
The exports wildcard "./components/*" only matches one path segment — actions/button is two,
so it looks for dist/components/actions/button/ which doesn't exist 
(the old Webpack build produced nested actions/button/ directories;
 the new Vite/Rolldown build flattens them).

To fix (when you want to code):

Add explicit aliases in solid-ui's exports mapping the old nested paths to the flat dist
Or add an "./components/*/*" export pattern
Or update profile-pane imports to flat paths like solid-ui/components/button
```

SharonStrats:
I am working on this PR https://github.com/SolidOS/profile-pane/pull/402 to update profile-pane to use the new buttons. It updates the import as well.

bourgeoa:
For compatibility with existing SolidOS panes and other existing users of solid-ui I feel we should keep nested path versus the new flat path.

NoelDeMartin:
Hey @bourgeoa, thanks for taking a look.

> I feel that that that changing to include solid-logic and/or rdflib must be on both ESM and legacy.

We're only building the UMD bundle for backwards compatibility, that's why I tried to leave it as unchanged as possible... But yes, if it weren't a deprecated CDN, I would agree with you.

In any case, if you still want to build it without externalizing the dependencies it's also ok with me, I just don't see any advantage.

> Not knowing how to commit dependencies updates/removed I created #794

Ok, let's look at that PR after we merge this to keep this PR from getting any more complicated. You should be able to rebase/cherry-pick once this is merged to have a clean PR.

> I tried to build profile-pane with solid-ui and got 37 errors of same type

That is fixed in one of the two related PRs:

- https://github.com/SolidOS/mashlib/pull/419
- https://github.com/SolidOS/profile-pane/pull/398

It's possible that other panes have issues I haven't noticed, but I prefer that they break now so that we notice (in the staging branches, this change shouldn't affect production).

> For compatibility with existing SolidOS panes and other existing users of solid-ui I feel we should keep nested path versus the new flat path.

You mean the imports such as `solid-ui/components/actions/button`? As far as I know, those were added in the last release and weren't documented anywhere. They are part of the `v2` folder, which was a temporary folder that Sharon created in her recent work.

I think it should be fine to remove these. As I said, if something is broken we should fix it before release, but I prefer to see the errors.

Now, another conversation entirely is if we actually want to group components so that they are imported like that, and not all like `solid-ui/components/*`. But I think for now it's better to just have everything flat. We can discuss again before making the release if we want to change this or group components in some way.

---

I think that addresses all your comments, can we move forward with merging this? Your review is still pending, but I could merge this already because Timea approved it. I just resolved the conflicts with staging too.

bourgeoa (approved):

PR #789, opened 2026-06-08
Add icon-only button support

m5x5:
## Summary
- add explicit icon-only support to Button
- add styling for icon-only buttons
- add Storybook coverage

### DIFF
--- a/src/design-system/components/dialog-header/DialogHeader.styles.css
+++ b/src/design-system/components/dialog-header/DialogHeader.styles.css
@@ -4,7 +4,7 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
-  padding: 15px;
+  padding: 12px 15px;
   gap: 15px;
   background: var(--solid-ui-color-slate-50);
   border-bottom: 1px solid var(--solid-ui-color-slate-200);
--- a/src/design-system/components/dialog/Dialog.styles.css
+++ b/src/design-system/components/dialog/Dialog.styles.css
@@ -8,6 +8,13 @@ dialog {
 
   solid-ui-dialog-header solid-ui-button {
     color: var(--solid-ui-color-gray-600);
+
+    icon-lucide-x {
+      width: 22px;
+      height: 22px;
+      justify-content: center;
+      align-items: center;
+    }
   }
 }
 
--- a/src/design-system/components/button/Button.stories.ts
+++ b/src/design-system/components/button/Button.stories.ts
@@ -26,6 +26,7 @@ const meta = {
     rightIcon: 'None',
     disabled: false,
     loading: false,
+    iconOnly: false,
   },
   argTypes: {
     variant: {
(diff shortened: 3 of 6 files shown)


timea-solid (approved):

src/design-system/components/dialog-header/DialogHeader.styles.css
```diff
@@ -4,7 +4,7 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
-  padding: 15px;
+  padding: 12px 15px;
```
SharonStrats, line 7: Are we moving away from using the variables in mashlib light theme? There are spacing variables for 12px and 15px. Also, wondering why we don't use rem? Curious for my work now on source-pane to know whether I should be doing things differently.

src/design-system/components/dialog/Dialog.styles.css
```diff
@@ -8,6 +8,13 @@ dialog {
 
   solid-ui-dialog-header solid-ui-button {
     color: var(--solid-ui-color-gray-600);
+
+    icon-lucide-x {
+      width: 22px;
```
SharonStrats, line 13: wondering about the size? I need for instance a 16px icon for the source-pane header.

NoelDeMartin (changes requested):
I have taken a look at the PR, and I think it's not necessary because the current code already supports what this is trying to achieve.

Here's some details:

- I already created an "icon only" configuration for buttons, which is basically using the "ghost" variant and including an `icon` slot. This is also showcased in the storybook with the Ghost variant, but this PR breaks it (please make sure to always check the storybook before changing any UI code).
- We *could* create an iconOnly variant or attribute in the Button component... But that's a matter of whether we like it more or less than the current component API. I'm not against it if you have suggestions for a better API, but I don't think it should change the implementation (unless we find some reason to).
- The iconOnly implementation from this PR creates a button that is 42x42px. That goes against [the current design](https://www.figma.com/design/88xQawSSXzfQprEZPaKplZ/Solid-OS?node-id=1101-14022&t=TugJatA98WG6pihm-4), which produces a 35x35px button (the one we currently have is 34x34px though, that should be fine for now). However, I know that 35x35 is too small for a clickable target... That's why the actual clickable area is 42x42px, configured with the `--solid-ui-clickable-area` variable.
- The PR also changes the close button from the modal, but again I think this was already fine in the current implementation (which is using the ghost variant + icon slot).

Did I miss something?

src/design-system/components/dialog/Dialog.styles.css
NoelDeMartin: The way icons are configured right now, you need to set their width/height like this yes. But I'd try to use variables whenever possible, instead of hard-coded px.

src/design-system/components/dialog-header/DialogHeader.styles.css
NoelDeMartin: Ideally yes, we should be using the variables from the design system css files (or primitive css files, which will probably be merged in the end).

timea-solid:
Thank you to reviewers. We will close this PR since work already has been done towards this.

PR #794, opened 2026-06-14
780 vite update dependencies

bourgeoa:


timea-solid:
I am not sure what to make if this PR. It is containing work from other PRs?

NoelDeMartin:
@timea-solid Yes, it was built from my working branch in #787. The commits are displayed as new commits because I rewrote history addressing some feedback before merging the PR. I guess @bourgeoa can cherry-pick his latest commit on top of main now, or reapply the changes, etc. But the last commit in the PR is what we should be looking at. I'll wait until he resolves the conflicts before adding my review though.

timea-solid (approved):

bourgeoa:
@NoelDeMartin 
rebase done
local failing test resolved test/unit/login/login.test.ts
build fails on solid-logic as expected

NoelDeMartin:
@bourgeoa The rebase didn't work as expected, because you've rebased the branch with my previous commits. If you look at the PR, it says there are 14 new commits but there should only be two (or maybe one, I think I already fixed that test in #796, I guess whichever gets merged first wins xD). 

Instead of a rebase, you should cherry-pick or apply your changes from a clean staging checkout. If you have issues doing that let me know and we can jump on a call and I help you do it.

bourgeoa:
@NoelDeMartin 
The issue was with `git push`
```
git fetch origin
git reset --hard origin/staging
git cherry-pick fe7f0f57ea468ab7c1494aeaf166b47aeb47d9e4
git push --force-with-lease
```
I had to fix test test/unit/login/login.test.ts

src/types/declarations.d.ts
```diff
@@ -2,3 +2,11 @@ declare module '*.sparql' {
   const content: string
   export default content
 }
+
+// unplugin-icons virtual modules — side-effect imports register custom elements
+declare module '~icons/lucide/*'
+declare module '~icons/svg-spinners/*'
+declare module '~icons/*'
```
NoelDeMartin: Instead of doing this, we can add the following line to `src/lib/env.d.ts`:

```ts
/// <reference types="unplugin-icons/types/web-components" />
```

However, I don't think it's necessary, given that icons are imported as side-effects so we don't need to use the result for anything. Did you get any errors that made you add this?

In any case, I don't think it's harmful either, so go ahead and add it if you want. But let's use the official types, rather than declaring it by hand.

src/types/declarations.d.ts
```diff
@@ -2,3 +2,11 @@ declare module '*.sparql' {
...
+
+// CommonJS `module` global used by iconBase.ts for script URI detection
+declare var module: { scriptURI?: string } | undefined
```
NoelDeMartin, line 7: Is this necessary? The only place I've found this being used is something that is casting `module` as any. Which isn't great either, but since this PR doesn't remove the `any`, it doesn't fix anything anyways.

In any case, the file using this (`src/lib/iconBase.ts`) should disappear once we migrate everything to unplugin-icons, so I think it'd be fine leaving the anys for now.

package.json
```diff
@@ -87,65 +87,61 @@
 		"mime-types": "^3.0.2",
 		"pane-registry": "^3.1.1",
 		"solid-namespace": "^0.5.4",
-		"tailwindcss": "^4.3.0",
+		"tailwindcss": "^4.3.1",
```
NoelDeMartin, line 90: We shouldn't update the tailwind dependency because we have a patch for the 4.3.0 version. In fact, I think we should update this to lock the dependency to `4.3.0`, rather than allowing any 4.x.

package.json
```diff
@@ -87,65 +87,61 @@
...
 		"neostandard": "^0.13.0",
 		"nock": "^15.0.0",
 		"patch-package": "^8.0.1",
 		"path-browserify": "^1.0.1",
 		"postcss": "^8.5.15",
-		"react": "^17.0.2",
-		"react-dom": "^17.0.2",
-		"react-is": "^17.0.2",
-		"storybook": "10.4.2",
+		"react": "^19.2.7",
```
NoelDeMartin: I wonder why we even have React as a dependency 🤔. If it's because of storybook, maybe it could be installed as a transitive dependency instead? I guess we should be able to remove it from here.

package.json
```diff
@@ -87,65 +87,61 @@
...
+		"react-dom": "^19.2.7",
+		"react-is": "^19.2.7",
+		"storybook": "10.4.4",
 		"typedoc": "^0.28.19",
-		"typescript": "^5.9.3",
+		"typescript": "^6.0.3",
```
NoelDeMartin, line 136: Any reason why we're upgrading the Typescript version? I'm all for it, but this may not be a trivial change, given that it's a new major version. Did you check if there are any breaking changes or something that could affect us?

NoelDeMartin (changes requested):
Overall I think it's fine to merge this after addressing a couple of comments I added, just some minor things about dependencies.

src/types/declarations.d.ts
bourgeoa: This was needed for typescript 6

src/types/declarations.d.ts
bourgeoa: The errror was
```
src/design-system/components/account/Account.ts:15:8 - error TS2882: Cannot find module or type declarations for side-effect import of '~icons/lucide/user'.

15 import '~icons/lucide/user'
          ~~~~~~~~~~~~~~~~~~~~
```
your solution works. Implemented

package.json
bourgeoa: Where is your patch (for information.
I generally don't agree to lock dependencies.
We should see tests failing when upgrading.

NoelDeMartin: The patch is in `patches/tailwindcss+4.3.0.patch`, it's applied using [patch-package](https://www.npmjs.com/package/patch-package).

You can upgrade the dependency if you want, but we should then update the patch to say 4.3.1 and make sure it's applied correctly.

In this case, this dependency is just importing a reset CSS file, so I think it should be pretty safe to lock to a particular version. This is part of the issues with dependabot and all the automated vulnerability checks... Maybe TailwindCSS will have some vulnerability reported some day, but I doubt our use of the library will be affected. I like this post by Dan Abramov talking about it: [npm audit: Broken by Design](https://overreacted.io/npm-audit-broken-by-design/).

package.json
bourgeoa: The reason for updating dependencies: you introduces a big change with vite.
The other reason is that in the past we were enable to cope with the flow of security dependency updates.
With AI actually it is worse.

I always `npm run prepublishOnly` that should contains :
- all builds
- all tests
- lint and typecheck
- eventually doc, storybook, ....

package.json
NoelDeMartin: Also, about seeing tests fail when upgrading... as I mentioned, this dependency is only used for CSS. It's very difficult to catch CSS regressions in tests, unless we do [visual snapshot testing](https://playwright.dev/docs/test-snapshots) or something like that, and I don't think we're ready to do that yet (we don't even have E2E tests).

package.json
NoelDeMartin: Let's talk about this in tomorrow's weekly meeting.

package.json
timea-solid: React was for storybook yes, but just because we had an old storybook. Now I am nto sure if we need it anymore.

NoelDeMartin:
I'm closing this for cleanup, given that it has diverged a lot from staging. We're going to have a meeting about CI and dependencies on Friday, so if anything needs to change we can just open a new PR.

PR #796, opened 2026-06-15
#795 Fix staging CI

### LINKED ISSUE #795: CI checks aren't run against staging
This means the only time CI checks are going to run is before doing a release, which isn't ideal. We should be running CI checks in every development PR as well. However, this is now not possible because we haven't released the development versions of dependencies (such as `solid-logic`). As a result, the current CI for staging is broken: https://github.com/SolidOS/solid-ui/actions/runs/27531922777

PS: unrelated to this, we should also be running `typecheck` and `typecheck-test` checks on CI.

NoelDeMartin:
This PR fixes the issues we have in CI as described in #795.

The fix clones the `solid-logic` dependency, and installs it locally before running the rest of the CI. This only works with `solid-logic` now, but this script could be refactored to take a list of dependencies that should be symlinked this way. Also, we could potentially use the same script in other repositories in the organization in order to run their respective CIs.

Overall, this is a working solution but it isn't ideal. Even though the CI works, developers will still need to manage their own local dependencies. Maybe this isn't too much of a problem, given that the expected development workflow is to use the Lerna config from the `solidos` repository.

If we want to improve this in the future, I can only think of two solutions:

- Release development versions of the libraries (some people have raised concerns about this approach in https://github.com/SolidOS/mashlib/issues/317).
- Migrate everything to a monorepo (would be nice, but it's obviously a lot of work).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - main
+      - staging
   pull_request:
     branches:
       - main
+      - staging
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm ci
+      - run: scripts/setup-ci.sh
       - run: npm test
       - run: npm run build
       - run: npm run doc
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
 		"lint-fix": "eslint --fix",
 		"typecheck": "tsc --noEmit",
 		"typecheck-test": "tsc --noEmit -p tsconfig.test.json",
-		"test": "vitest run",
+		"test": "npm run typecheck-test && vitest run",
 		"test-coverage": "vitest run --coverage",
 		"test-debug": "vitest --inspect-brk --no-file-parallelism",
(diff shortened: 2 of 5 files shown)


SharonStrats:
That's cool, it does get a bit more complicated though like solid-panes for instance it depends on all the panes. I wonder if it would get to be a bit much in that case.

NoelDeMartin:
> That's cool, it does get a bit more complicated though like solid-panes for instance it depends on all the panes. I wonder if it would get to be a bit much in that case.

The script shouldn't get over-complicated, it would just take a list of dependencies rather than assume that `solid-logic` is the only dependency. But yes, it's not super straight-forward... But I can't think of any better way to do it (besides the 2 long time solutions I mentioned).

timea-solid:
I am now working locally to test and build an entire staging of SolidOS in order to be able to test the Authn switch.
I was not expecting to have so many problems on staging, a branch which we agreed should always be green and stable and ready to release. This is because we do not have 'CI green' on the branches because the repos are depending on each other, and without a release this is getting very, very complicated.

While the script above is a nice solution to have, it is not realistic in say solid-panes where we need to pull in all repos.

For these reasons, I would propose to go back to releasing when pushing on the staging of all repos.

NoelDeMartin:
I'm closing this PR, as it doesn't seem like we'll be moving forward with this. Can reopen later if we decide to do it again, but looking at the comments in #795 it looks like we'll move forward with the prereleases approach instead.

I have opened a separate PR that just fixes the login tests, though: #803.

PR #798, opened 2026-06-16
Initial RDF forms component

### LINKED ISSUE #304: Create a starting RDF forms Web Component with an input field example
We will be creating a new Web Component based on an existing input web component but adding the rdf capabilities on it.

timea-solid:
This is for now:
* adding an overall RDF forms component to storybook
* testing its behaviour in Storybook by implementing a first RDF input Web Component


src/types/custom-elements.d.ts
```diff
@@ -0,0 +1,5 @@
+declare global {
+  interface HTMLElementTagNameMap {
```
NoelDeMartin, line 30: I like this, but we should generate it automatically at build time instead of doing it by hand. We could create a new Vite plugin that emits this file in `vite-config/components.ts`.

src/components/rdf-form/RDForm.stories.ts
```diff
@@ -0,0 +1,115 @@
...
+const render = defineStoryRender<typeof meta.argTypes>(({ rdfTurtleFormatSource, rdfURI, whichForm, rdfName, subjectTurtleFormatSource, subjectName, subjectURI }) => {
+  return html`
+        <solid-ui-rdf-form 
+          rdfTurtleFormatSource=${rdfTurtleFormatSource}
+          rdfURI=${rdfURI}
+          whichForm=${whichForm}
+          rdfName=${rdfName}
+          subjectTurtleFormatSource=${subjectTurtleFormatSource}
+          subjectName=${subjectName}
+          subjectURI=${subjectURI}>
```
NoelDeMartin: Do we need all these attributes? This is a bit confusing, I'm not sure what each property does and why we need so many.

Could it work with something like this?

```html
<solid-ui-rdf-form
    form="https://solidos.solidcommunity.net/public/2021/solidUiFormTestData/dummyFormTestFile.ttl#form"
    subject="https://solidos.solidcommunity.net/public/2021/alice.ttl#me"
>
</solid-ui-rdf-form>
```

Internally, it could resolve the form and subject from the store associated to the authenticated user.

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,137 @@
...
+    accessor whichForm = 'this'
+
+    @property({ type: String })
+    accessor rdfTurtleFormatSource = ''
+
+    @property({ type: String })
+    accessor rdfName = ''
+
+    @property({ type: String })
+    set rdfURI (value: string) {
```
NoelDeMartin: I think these two propertie (rdfURI and subjectURI) can probably be simplified a lot, I'm not sure we need the `_parsedUrl` variables or duplicating the URL parsing code. Check out [Lit converters](https://lit.dev/docs/components/properties/#conversion-converter).

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,137 @@
...
+      }
+    }
+
+    get subjectURI (): string {
+      return this._parsedUrl2 ? this._parsedUrl2.href : ''
+    }
+
+    render () {
+      // TODO: detect format
+      loadDocument(store, this.rdfTurtleFormatSource, this.rdfName, this.rdfURI) // load form
```
NoelDeMartin: Instead of using solid-logic's store directly, maybe we can encapsulate this into the `Auth` interface?

Not sure if we want to couple both concepts though (store and Auth), but maybe we could given that they are already coupled in the solid-logic implementation (I think).

This would also allow us to improve the storybook stories so that they don't require solid-logic. Right now, it seems like we're passing the full RDF graph to the component, but that shouldn't be necessary (I suspect that's a workaround because solid-logic doesn't work in storybook).

Check out how the authentication is configured in storybook and elsewhere, and make sure to learn about [Lit context](https://lit.dev/docs/data/context/).

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,137 @@
...
+        const value = typeNode ? ((typeNode as any).value || String(typeNode)) : ((item as any).value || String(item))
+        const hashIndex = value.lastIndexOf('#')
+        return {
+          value: item,
+          fieldValue: hashIndex >= 0 ? value.slice(hashIndex + 1) : value
+        }
+      })
+      const me = Namespace(this.subjectURI + '#')(this.whichSubject)
+
+      return html`
```
NoelDeMartin, line 93: Other than the inputs, shouldn't this output a native `<form>` as well? Ideally, whatever is generated from this component ends up being a "real" form. That is important in order to integrate with accessibility tooling and user expectations ([implicit form submission](https://www.w3.org/TR/2018/SPSD-html5-20180327/forms.html#implicit-submission), etc.).

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,137 @@
...
+            case 'IntegerField':
+            case 'DecimalField':
+            case 'FloatField':
+            case 'TextField':
+            case 'SingleLineTextField':
+            case 'NamedNodeURIField': {
+              return html` <solid-ui-rdf-input 
+                .store=${store} 
+                .formSubject=${sym(part.value)} 
+                .dataSubject=${me}
```
NoelDeMartin: I think all inputs should also have a `name` attribute. Otherwise, they'll be ignored by the form (again, important for a11y, etc.).

src/components/rdf-input/RDFInput.ts
```diff
@@ -0,0 +1,100 @@
...
+  // example RDF Turtle format source:
+  // :nameField a ui:SingleLineTextField ;
+  //   ui:property vcard:fn;
+  //   ui:label "name" .
+
+  // formSubject describes the field metadata
+  // dataSubject points to the data resource containing the value
+
+  @property({ attribute: false })
+  accessor store!: LiveStore
```
NoelDeMartin: Instead of passing the store as a property, I think we should use Lit context instead. See my other comments talking about this.

src/components/rdf-input/RDFInput.ts
```diff
@@ -0,0 +1,100 @@
...
+    const fieldType = this.formSubject ? mostSpecificClassURI(this.store, this.formSubject) : undefined
+    const params = fieldType ? fieldTypeParams[fieldType] ?? {} : {}
+    const inputType: InputType = params.type ?? 'text'
+
+    // for populating the HTML input element
+    const selectedTerm = this.getSelectedTerm(this.dataSubject, uiPropertyTerm, this.formSubject, params)
+    const inputValue = this.termToInputValue(selectedTerm, params)
+
+    return html`
+      ${inputLabel ? html`<label>${inputLabel}</label>` : ''}
```
NoelDeMartin: This label won't be associated with the input, given that it doesn't have a `for` attribute. One way to avoid using `for` is to include both the label text and the <input> inside the label element, like this:

```html
<label>
    Name:
    <input name="name">
</label>
```

I know it looks counterintuitive, but I'm pretty sure this is correct HTML and I've used it many times :).

In any case, the `for` attribute is probably the best solution. We can use `generateId()` from `src/lib/components/ids.ts` to create a unique id for each input.

src/lib/forms/fieldParams.ts
```diff
@@ -0,0 +1,140 @@
...
+export type FieldParamsObject = {
+  size?: number, // input element size attribute
+  type?: InputType, // input element type attribute. Default: 'text' (not for Comment and Heading)
+  element?: string, // element type to use (Comment and Heading only)
+  style?: string, // style to use
+  dt?: string, // xsd data type for the RDF Literal corresponding to the field value. Default: xsd:string
+  defaultInputValue?: string, // e.g. 'mailto:'. Default value in input field, will be removed when displaying actual value to user.
+  namedNode?: boolean, // if true, field value corresponds to the URI of an RDF NamedNode. Overrides dt and defaultInputValue.
+  pattern?: RegExp // for client-side input validation; field will go red if violated, green if ok
+}
```
NoelDeMartin, line 38: Many of these are currently unused, right? I see that this is a WIP, so it's fine. But before merging, I would remove everything that isn't used (you know, YAGNI, removing dead code, etc.).

Also, I'm not sure I like the `style` config... that goes a bit against the idea of RDF forms, given that different UI libraries may render forms differently, and inline styles are probably not going to work correctly in all of them. But if that was a design decision when creating the RDF forms, ok.

NoelDeMartin (commented):
Added a few initial comments, mostly regarding a11y and component architecture.

timea-solid:
RDFInput will make use of the new Input components when it is merged: https://github.com/SolidOS/solid-ui/pull/815

timea-solid:
We need to improve the name property of the generated form and make sure they are unique.
@NoelDeMartin do you have a suggestion for this?

All HTML elements created automatically from the ontology ui file need names but there could be more input fields called input-name. I am not sure I can just use a random id generator.

src/components/input/Input.styles.css
```diff
@@ -23,5 +23,12 @@
             color: var(--solid-ui-color-gray-700);
             font-size: inherit;
         }
+
+        input:read-only {
+            border: none;
+            padding: 0;
+            cursor: pointer;
```
NoelDeMartin:
```suggestion
            cursor: not-allowed;
```

I guess this is more appropriate? Maybe we should add a story in the input storybook to showcase this state as well, even though it's not super common so maybe it isn't necessary.

src/lib/forms/store/RDFFormsStore.ts
```diff
@@ -0,0 +1,8 @@
+import { LiveStore } from 'rdflib'
+import { StoreContext } from './StoreContext'
+
+export default class RDFFormsStore implements StoreContext {
```
NoelDeMartin: I think this should be named differently, to make it obvious that this is a stub implementation. I am calling the auth `NoopAuth`, so maybe NoopStore? Also, as we discussed the store is not necessarily used only in forms, so we should avoid mentioning anything about forms in the error messages (the form components can override the error handling to give a more specific error if necessary).

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,205 @@
...
+import { sym, Namespace, LiveStore } from 'rdflib'
+import '@/components/rdf-input'
+import { DEFAULT_STORE, storeContext, StoreContext } from '@/lib/forms/store/StoreContext'
+
+@customElement('solid-ui-rdf-form')
+export default class RDFForm extends WebComponent {
+    @consume({ context: storeContext, subscribe: true })
+    private accessor storeContext: StoreContext = DEFAULT_STORE
+
+    @property({ attribute: false })
```
NoelDeMartin, line 36: is the `attribute: false` necessary? I guess we do want to observe is this property changes, although it'll be very uncommon (wouldn't even worry about testing it for now).

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,205 @@
...
+    accessor passedInStore: LiveStore | null = null
+
+    private get currentStoreContext (): StoreContext | null {
+      if (this.passedInStore) {
+        return { store: this.passedInStore }
+      }
+
+      return this.storeContext !== DEFAULT_STORE ? this.storeContext : null
```
NoelDeMartin: Instead of returning null, I think this should return the default store. The point of having a "noop" store is that it already throws an error when we attempt to use it, and it should never happen anyways. By returning null here, we're forced to handle the "null" everywhere else in the component (which I see is handled by converting the error into a warning, IMO it should be an error).

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,205 @@
...
+    private accessor failed: boolean = false
+
+    @state()
+    private accessor submitting: boolean = false
+
+     @state()
+    private accessor entireDataIsReadonly: boolean = true
+
+    @state()
+    private accessor _parsedUrl: URL | null = null
```
NoelDeMartin: I mentioned this in my previous review, but as reminder I don't think we need these two attributes, see [Lit converters](https://lit.dev/docs/components/properties/#conversion-converter).

src/components/rdf-form/RDForm.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+    rdfName: { control: 'text' },
+    subjectTurtleFormatSource: { control: 'text' },
+    subjectName: { control: 'text' },
+    subjectURI: { control: 'text' }
+  },
+} as const
+
+const render = defineStoryRender<typeof meta.argTypes>(({ rdfTurtleFormatSource, rdfURI, whichForm, rdfName, subjectTurtleFormatSource, subjectName, subjectURI }) => {
+  return html`
+      <storybook-provider>
```
NoelDeMartin: The `defineStoryRender` method already wraps the story in a `<storybook-provider>`, no need to duplicate it here.

src/components/rdf-input/RDFInput.ts
```diff
@@ -0,0 +1,200 @@
...
+  // example RDF Turtle format source:
+  // :nameField a ui:SingleLineTextField ;
+  //   ui:property vcard:fn;
+  //   ui:label "name" .
+
+  // formSubject describes the field metadata
+  // dataSubject points to the data resource containing the value
+
+  @consume({ context: storeContext, subscribe: true })
+  private accessor storeContext: StoreContext = DEFAULT_STORE
```
NoelDeMartin, line 24: The parent RDFForm accepts a store as an attribute, but it seems like child Inputs would ignore it completely and fall back to the global store? Maybe the store in form controls should be read from a "form" context provided by the RDFForm component (and yes, that means that using an RDFInput outside of a RDFForm will throw an error).

src/types/custom-elements.d.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * This file is auto-generated by vite-config/components.ts.
```
NoelDeMartin: If this file is auto-generated, it should be included in `.gitignore`.

vite.config.ts
```diff
@@ -6,9 +6,11 @@ import babel from './vite-config/babel'
 import css from './vite-config/css'
 import icons from './vite-config/icons'
 import { cdnLegacyConfig, cdnConfig } from './vite-config/cdn'
-import { discoverComponents } from './vite-config/components'
+import { discoverComponents, generateCustomElementsTypes } from './vite-config/components'
 import { stylesConfig } from './vite-config/styles'
 
+generateCustomElementsTypes()
```
NoelDeMartin: This is not implemented properly. The way this works is that whenever this file is executed, the declaration file will be written to disk... Instead, it should be invoked as part of Vite lifecycle. You can take a look at the [unplugin-vue-components](https://github.com/unplugin/unplugin-vue-components#typescript) Vite plugin to see how to do this properly, given that it does something very similar.

Also, instead of declaring just the form components, I think we could declare all the components in the `src/components` folder.

src/components/rdf-form/RDFForm.ts
timea-solid: sone in commit: https://github.com/SolidOS/solid-ui/pull/798/commits/9d9f48abf505e19811643f4368118d3e39ec2304

NoelDeMartin:
I just saw this now, here's my thoughts:

> We need to improve the name property of the generated form and make sure they are unique.
@NoelDeMartin do you have a suggestion for this?
>
> All HTML elements created automatically from the ontology ui file need names but there could be more input fields called input-name. I am not sure I can just use a random id generator.

It depends on how the form is going to be used by the consumer. If everything is handled in javascript, there's nothing wrong with using names generated by the id generator. However, if the consumer wants to actually submit this form to a server they would need to know the actual names.

I haven't looked at the `ui:` ontology, but does it have an optional field for the name of the form control? If it doesn't, we should add it.

Basically, the behaviour I would like to see is that by default it generates random input names using the id generator, but if the RDF definitions specify the name of the form control, we'll use that. If those are not unique, I don't think we should do anything about it, that's the responsibility of whoever is defining the RDF (if they want two fields to have the same name, let them do it, although we could throw a helpful warning in the console or something).

That would be useful for different reasons, not just to handle the form submission. For example, the name of an input can influence some browser extensions or autocompletes.

timea-solid:
What we also need to decide is if we take the input size property into consideration. 
How is the input length determined in the new Input Web Component?

NoelDeMartin:
I guess this is something to discuss, but personally I don't think the RDF should be able to affect the UI (things like size). That's the whole point of the approach, that the form gets rendered using the RDF definition in a way that works for the current UI/project.

I guess it's fine to have some properties like "primary color" for theming or something like that. But even then, shouldn't the "primary color" be set by the application where the form is rendered in? Maybe it can be useful to change an accent or something, to distinguish a button or input from others... But I'm not sure I like the idea.

As per how the size is determined now, I haven't looked too much into it, but usually inputs have a min-width or take 100% of their container. That's something that should be configured in the form. For now I think it's fine to simply have a max-width in the form and otherwise take up the entire container. We can tweak these things later, possibly with some properties in the component.

timea-solid:
> I just saw this now, here's my thoughts:
> 
> > We need to improve the name property of the generated form and make sure they are unique.
> > @NoelDeMartin do you have a suggestion for this?
> > All HTML elements created automatically from the ontology ui file need names but there could be more input fields called input-name. I am not sure I can just use a random id generator.
> 
> It depends on how the form is going to be used by the consumer. If everything is handled in javascript, there's nothing wrong with using names generated by the id generator. However, if the consumer wants to actually submit this form to a server they would need to know the actual names.
> 
> I haven't looked at the `ui:` ontology, but does it have an optional field for the name of the form control? If it doesn't, we should add it.
> 
> Basically, the behaviour I would like to see is that by default it generates random input names using the id generator, but if the RDF definitions specify the name of the form control, we'll use that. If those are not unique, I don't think we should do anything about it, that's the responsibility of whoever is defining the RDF (if they want two fields to have the same name, let them do it, although we could throw a helpful warning in the console or something).
> 
> That would be useful for different reasons, not just to handle the form submission. For example, the name of an input can influence some browser extensions or autocompletes.

I have decided to do a generateId (random id) since we did not have the name attribute at all in the past. A random one is enough for now.

src/components/rdf-input/RDFInput.ts
```diff
@@ -0,0 +1,195 @@
...
+  @property({ type: Boolean, reflect: true })
+  accessor readonly = false;
+
+  render () {
+    const document = this.getDocument(this.formSubject)
+
+    // for building the HTML input element
+    const uiPropertyTerm = this.getFormProperty(this.formSubject, ns.ui('property'), document)
+    const inputLabel = this.getInputLabel(this.formSubject, uiPropertyTerm, document)
+    const readonly = this.getReadOnly(this.readonly, this.formSubject, document)
```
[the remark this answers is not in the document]
timea-solid: done

src/components/rdf-input/RDFInput.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    // for populating the HTML input element
+    const selectedTerm = this.getSelectedTerm(this.dataSubject, uiPropertyTerm, this.formSubject, params)
+    const placeholder = readonly ? '' : this.defaultInputValue(params)
+    const inputValue = this.termToInputValue(selectedTerm)
+
+    return html`
+      <solid-ui-input
+        label="${inputLabel}"
+        name="name-${generateId()}"
```
NoelDeMartin: The `generateId()` helper shouldn't be called inside of `render`, if we do we'll get a different ID every time the component re-renders. Instead, we should set it on a private property in the constructor.

Actually, I see this component is not using the `FormControl` trait. I wonder if it should? If it does, that trait already generates an internal id for each control, so we'll be able to use that one to get the name.

src/components/rdf-input/RDFInput.ts
```diff
@@ -0,0 +1,200 @@
...
+  //   ui:label "name" .
+
+  // formSubject describes the field metadata
+  // dataSubject points to the data resource containing the value
+
+  @consume({ context: storeContext, subscribe: true })
+  private accessor storeContext: StoreContext = DEFAULT_STORE
+
+  @property({ attribute: false, type: Object })
+  accessor formSubject!: NamedNode
```
NoelDeMartin:
```suggestion
  accessor formSubject: NamedNode | null = null
```

It's generally an anti-pattern to use `!` in type declarations. That assumes that the component will always be used with the "required" attributes, but what if someone uses it without them? I'd rather we control this use-case in the code to throw an appropriate error message, rather than something like "trying to call .something in `undefined`". We can encapsulate reading this property in a getter, so it shouldn't be much of a change.

Same feedback to other attributes in this and other components.

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,172 @@
...
+      return this.passedInStore
+        ? ({ store: this.passedInStore } as StoreContext)
+        : this.storeContext
+    }
+
+    @state()
+    private accessor entireDataIsReadonly: boolean = true // to protect data, we default to not editable
+
+    @state()
+    private accessor _loadVersion = 0
```
NoelDeMartin, line 47: I'm not sure what this and the `storeVersion` proeprties are for, can you help me understand?

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,172 @@
...
+    accessor passedInStore: LiveStore | null = null
+
+    private get currentStoreContext (): StoreContext {
+      return this.passedInStore
+        ? ({ store: this.passedInStore } as StoreContext)
+        : this.storeContext
+    }
+
+    @state()
+    private accessor entireDataIsReadonly: boolean = true // to protect data, we default to not editable
```
NoelDeMartin, line 44: This seems to be used only in the `render()` method, I think it'd be better to keep it as a function-scoped variable rather than component state.

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,172 @@
...
+                  return html`<input .rdf=${part}></input>`
+              default:
+                  return html`<div>Unknown part type: ${part}</div>`
+          }
+        })}
+      </form>
+    `
+    }
+
+    protected updated (changedProperties: Map<PropertyKey, unknown>) {
```
NoelDeMartin: I think this should override `willUpdate` rather than `update`. It shouldn't make much of a difference, but it's more appropriate.

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,172 @@
...
+
+@customElement('solid-ui-rdf-form')
+export default class RDFForm extends WebComponent {
+    @consume({ context: storeContext, subscribe: true })
+    private accessor storeContext: StoreContext = DEFAULT_STORE
+
+    @property({ attribute: false })
+    accessor passedInStore: LiveStore | null = null
+
+    private get currentStoreContext (): StoreContext {
```
NoelDeMartin: I think this getter is only used to get the store. It could return the store directly to avoid having to call `.store` everywhere.

src/components/rdf-form/RDFForm.ts
```diff
@@ -0,0 +1,172 @@
...
+              case 'TextField':
+              case 'SingleLineTextField':
+              case 'NamedNodeURIField': {
+                return html` <solid-ui-rdf-input 
+                  .formSubject=${sym(part.value)} 
+                  .dataSubject=${sym(subjectUrl)}
+                  .storeVersion=${this._loadVersion}
+                  .readonly=${this.entireDataIsReadonly}
+                ></solid-ui-rdf-input>
+                <br>`
```
NoelDeMartin, line 116: I don't think this `<br>` should be necessary, we should deal with these in CSS.

src/components/select/Select.ts
```diff
@@ -48,6 +51,7 @@ export default class Select extends WebComponent {
           id="${this.inputTrait.inputId}"
           name=${this.name}
           ?required=${this.required}
+          ?disabled=${this.readonly}
```
NoelDeMartin: I think native selects also have a `readonly` attribute, so maybe we should use that one instead of `disabled`.

src/storybook/components/StorybookProvider.ts
```diff
@@ -28,6 +33,10 @@ export class StorybookProvider extends WebComponent {
     }
 
     this.auth.account = new Account(this.webId, this.avatarUrl)
+
+    if (this.store) {
+      // read `store` so the property is considered used
+    }
```
NoelDeMartin, line 44: Was this added to silence a typescript or linting error? If that's the case, I think it'd be better to add an ignore-comment rather than add some runtime behaviour.

.gitignore
```diff
@@ -1,6 +1,7 @@
 node_modules
 dist
 src/versionInfo.ts
+src/types/custom-elements.d.ts
```
NoelDeMartin, line 4: Since this was added to the gitignore (which I like!), we should remove the file from the repository as well.

vite-config/components.ts
```diff
@@ -20,3 +22,56 @@ export function discoverComponents(): string[] {
...
+export function generateCustomElementsTypes(): void {
+    const components = discoverComponents()
+
+    const lines = [
+        '/**',
+        ' * This file is auto-generated by vite-config/components.ts.',
+        ' * Do not edit this file directly.',
+        ' */',
+        '',
+    ]
```
NoelDeMartin: I don't think this is necessary anymore, given that this generated file won't be tracked in the repository anymore.

NoelDeMartin (changes requested):
I spotted a couple of things that can be improved (some actual problems, like the `generateId()` in the render), but otherwise I think this is pretty close to being ready.

timea-solid:
Implemented all feedback that made sense.

PR #799, opened 2026-06-16
#790 Refactor components organization

### LINKED ISSUE #774: Export new Web Components

### LINKED ISSUE #790: Restructure Web Components
After some team discussions, we've decided that we won't differentiate web components by "primitives" or "design system" anymore. Instead, they will all be plain web components exported by the `solid-ui` package; some more reusable or ambitious than others.

We can still categorize them in some way for code organization or documentation, but they won't be separated in a different package or namespace (they'll all have the `<solid-ui` prefix).

NoelDeMartin:
- Move all components to src/components
- Moved common component styles to WebComponent base class
- Renamed styles.css to theme.css
- Added @/ imports and aliases
- Export component utils


src/storybook/auth/StorybookAuth.ts
```diff
@@ -1,5 +1,4 @@
-import { AuthContext } from '../../primitives/lib/auth/context'
-import Account from '../../primitives/lib/auth/Account'
+import { Account, AuthContext } from '@/lib/auth'
 
 export default class StorybookAuth implements AuthContext {
```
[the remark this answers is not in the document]
NoelDeMartin: I'm not sure if Copilot hallucinated this or it's a real issue, but I haven't seen any linting or build errors for this so it should be fine. The `callback` parameter is not used in the `StorybookAuth`, so there's no point in declaring it. That doesn't break the interface, given that passing additional parameters at runtime is fine, they'll just get ignored.

package.json
```diff
@@ -87,7 +87,7 @@
 		"mime-types": "^3.0.2",
 		"pane-registry": "^3.1.1",
 		"solid-namespace": "^0.5.4",
-		"tailwindcss": "^4.3.0",
+		"tailwindcss": "4.3.0",
```
timea-solid, line 112: Just out of curiosity: this has to be a fixed version?

timea-solid (approved):

package.json
NoelDeMartin: Yes, I discussed this with Alain here: https://github.com/SolidOS/solid-ui/pull/794#discussion_r3418775419

PR #801, opened 2026-06-17
Fix auth test

timea-solid:
Why this fails
`vi.spyOn(authSession, 'info', 'get') i`s not defined as a getter property on that object.

`vi.spyOn(obj, prop, 'get' `requires prop to be an accessor with a getter descriptor.
In solid-logic is a plain property on the session object, not a getter.
So Vitest throws: `[The property "info" is not defined on the object.]`

Find fix I chose in this PR.


timea-solid:
ficed in https://github.com/SolidOS/solid-ui/pull/803

PR #812, opened 2026-06-22
CI tweaks + prerelease 3.1.3-4

NoelDeMartin:


timea-solid (approved):

package.json
```diff
@@ -21,25 +21,11 @@
...
-		},
-		"./components/forms/combobox": {
-			"types": "./dist/components/combobox/index.d.ts",
-			"import": "./dist/components/combobox/index.esm.js",
-			"require": "./dist/components/combobox/index.cjs.js"
-		},
-		"./components/media/photo-capture": {
-			"types": "./dist/components/photo-capture/index.d.ts",
-			"import": "./dist/components/photo-capture/index.esm.js",
-			"require": "./dist/components/photo-capture/index.cjs.js"
```
bourgeoa, line 42: This is an example of losing compatibility. Here with current dev mashlib package.
This was introduced to remove mashlib build errors in the UVDSL test.

bourgeoa:
3.1.3-3 was already published in #811 
The current workflow shows guarding for staging branch as expected.
```
Run # Skip publishing when PR originates from staging (will publish after merge)
```
commit a new version to see it published with a dev tag after merging

package.json
NoelDeMartin: Ok, but the correct fix here would be to remove the legacy imports from any pane that is still using them, rather than adding them back.

NoelDeMartin:
>  The current workflow shows guarding for staging branch as expected.

I'm not sure it's fixed, because I'm pretty sure I opened the previous PR after the fix.

Looking at the code, I see you're comparing with `github.base_ref`. That is the target branch in a PR, that's why it's still being released on the PR (before merge). If you want to do it after merge, you should compare with `github.ref`. That will run it for every new commit in `staging`, regardless if it comes from a PR or a direct commit (which we should never do!).

PR #815, opened 2026-06-23
#773 Improve a11y + prepare components for contacts-pane

### LINKED ISSUE #773: Review a11y for design system and primitive Web Components
As part of #769, a bunch of Web Components have been implemented that may not be completely accessible. In particular, these could be the `Menu`, `Combobox`, and `Dialog` components.

Before looking into this, though, we still have to consider whether we can use 3rd party components from [Web Awesome](https://webawesome.com/) or other libraries. If we do, maybe we can end up removing all the new functionality that has been implemented without an a11y review.

To clarify, the reason why these weren't implemented with full a11y support from the beginning is because doing that requires a lot more work. I preferred to get a basic version going, and once we decide whether we're using 3rd party components or not we can improve them.

NoelDeMartin:
This PR includes a couple of things necessary for https://github.com/SolidOS/contacts-pane/issues/283, as well as improving a11y:

- Integrates Web Awesome for accessible menus (See #773).
- Implements traits for inputs and dialogs.
- Adds new form components: Input and Select.


NoelDeMartin:
CI is failing because the PR comes from a forked repository, I've been pushing my branches to the main repo to work around this issue until now, but I'm leaving this "broken" because it's not so urgent to merge this and we should solve that issue.

@bourgeoa can you take a look? I think the problem was introduced in your recent CI changes.

timea-solid (approved):
very interesting

PR #817, opened 2026-06-25
#769 Prepare utils for Header component

### LINKED ISSUE #769: Replace Header with Design System components

### LINKED ISSUE #771: Remove DESIGN_SYSTEM_HEADER_ACCOUNT feature toggle
Working on #769, I've introduced a feature flag to isolate the experimental functionality until it's finished. At that point, we can remove the feature toggle altogether and all the legacy code that the new functionality is replacing.

NoelDeMartin:
PR blocked by:

- https://github.com/SolidOS/toolkit/pull/5


src/lib/components/traits/InputTrait.ts
```diff
@@ -49,7 +51,7 @@ export default class InputTrait implements WebComponentTrait {
 
   renderLabel () {
     return this.target.label
-      ? html`<label for="${this.inputId}">${this.target.label}</label>`
+      ? html`<label id="${this.labelId}" for="${this.inputId}">${this.target.label}</label>`
```
timea-solid, line 54: I saw this in the other PR - don't the id and for need to be the same?

timea-solid (approved):
Nice to see it come together more and more like the menu items, the mobile breakpoint again.
Do you think this will solve the duplicated loading of components?

src/lib/components/traits/InputTrait.ts
NoelDeMartin: No, the "for" should be the id of the `<input>`, whilst labelId is the id for the `<label>`, they are different elements. The reason why we need an id in the label is that in some cases, we need to reference it with the `aria-labelledby` attribute. For example, in the combobox the listbox needs to reference it (which is a plain `<div>` with the "listbox" role). And we can't reuse the `for` attribute to target the listbox because we have 2 elements using the same label, the `<input>` and the `<div role="listbox">` (and ids have to be unique).

PR #818, opened 2026-06-30
remove duplicate load of button component

timea-solid:
The old v2 button was loading with the same tag name
TODO:
* remove the entire code under v2 carefully


NoelDeMartin (approved):
I'm ok with removing this, but if we really want to test if it fixes the deduplication we'll have to disable the conditional registration of the component as well.

Either way, I think it's ok to remove the dead code (this button shouldn't be used anymore), so I'm approving it.

PR #824, opened 2026-07-01
Atkinson font added here for now

### LINKED ISSUE #293: Font needs to be added
In order to follow the design.inrupt.org styles we need to add the font. Raleway and Roboto
Below is the link found in https://github.com/inrupt/solid-style-guide/blob/master/css/main.css
but we need to actually install it into our library versus referring to the link.

@import url("https://fonts.googleapis.com/css?family=Raleway:400,400i,700,700i|Roboto:300,300i,400,400i,700,700i");

timea-solid:
I added the font for now here.
However, it seems like it is overwritten by the --wa-font-family-body: ui-sans-serif, system-ui, sans-serif;

I also need to delete the font in mashlib light.css.


NoelDeMartin (commented):
This can work for now, and maybe we could even release it like this... But we should discuss whether we want to add a dependency with google for the fonts. People using Solid will be particularly privacy-sensitive, so I don't know if they'll like seeing that every time they open SolidOS, they'll be pinging google's servers.

As per the web awesome fonts variables, I think it should be fine to simply define `--wa-font-family-body` as well. See their [theming docs](https://webawesome.com/docs/customizing#customizing-with-css).

timea-solid:
Consider also merging https://github.com/SolidOS/mashlib/pull/441 (non blocking)

NoelDeMartin (approved):
As I mentioned, I'm ok with merging this as is... but before releasing a new version, we should double-check if we really want to be injecting Google's CDN into SolidOS.

PR #826, opened 2026-07-03
make an "add as friend" web component

### LINKED ISSUE #708: createAddMeToYourFriendsButton should become a web component in solid-ui
createAddMeToYourFriendsButton is used in profile-pane and imported in social-pane. This is not corect architecturarly.

SharonStrats:
This moves over the code from profile-pane for adding as a friend and creates a Web Component from the code.


timea-solid:
Looks ok to me, but I would appreciate Noels review.

src/components/button-add-friend/ButtonAddFriend.ts
```diff
@@ -0,0 +1,182 @@
...
+  private accessor auth: AuthContext = DEFAULT_AUTH_CONTEXT
+
+  @property({ type: Boolean })
+  accessor disabled: boolean | undefined = undefined
+
+  @property({ attribute: false })
+  accessor subject: NamedNode | undefined = undefined
+
+  @property({ attribute: false })
+  accessor context: DataBrowserContext | undefined = undefined
```
NoelDeMartin, line 36: We shouldn't be passing the context as an attribute, it should be a context provided by a parent component. We already do this with the `auth` context, so we could reuse that one or create a dedicated one for the store.

Timea has been working on this, and she's leaning more towards having a separate "StoreContext". Check it out in her PR, as they should both follow the same approach: https://github.com/SolidOS/solid-ui/pull/798

src/components/button-add-friend/ButtonAddFriend.styles.css
NoelDeMartin, line 1: Is there any reason why these styles override the default `<solid-ui-button>` styles? It'd be better if they are the same for consistency.

src/components/button-add-friend/ButtonAddFriend.stories.ts
NoelDeMartin, line 1: Looking at the code, I see that we also have a statusMessage that is rendered whenever something goes wrong with the operation. Ideally, we should also be able to see that in the storybook.

src/components/button-add-friend/ButtonAddFriend.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  @property({ attribute: false })
+  accessor buttonLabel = addMeToYourFriendsButtonText
```
NoelDeMartin: Maybe this should be state rather than a property... Or seeing that this is derived from other state, maybe it could just be a getter that we call in the render.

src/components/button-add-friend/ButtonAddFriend.ts
```diff
@@ -0,0 +1,182 @@
...
+  static styles = styles
+
+  @consume({ context: authContext, subscribe: true })
+  private accessor auth: AuthContext = DEFAULT_AUTH_CONTEXT
+
+  @property({ type: Boolean })
+  accessor disabled: boolean | undefined = undefined
+
+  @property({ attribute: false })
+  accessor subject: NamedNode | undefined = undefined
```
NoelDeMartin: Instead of taking a NamedNode, I think this should be a plain string and reflected in the DOM (with `attribute: true`, or simply not declaring it as that's the default). See how Timea is also handling the subject in the RDFForm component.

src/components/button-add-friend/ButtonAddFriend.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  @property({ attribute: false })
+  accessor context: DataBrowserContext | undefined = undefined
+
+  @property({ attribute: false })
+  accessor buttonLabel = addMeToYourFriendsButtonText
+
+  @property({ attribute: false })
+  accessor statusMessage = ''
```
NoelDeMartin: Again, this should probably be state, not a property.

NoelDeMartin (commented):
I think the general direction is ok, but there are a couple of things that should be improved to make this more "idiomatic".

Also, regarding the name, maybe it should be called "add-friend-button" instead? I worry this is too specific to be in the generic `solid-ui` package, but I understand we're doing this to use it in two different panes so I guess it's ok. Whenever we prepare developer documentation, we should probably make sure that this type of component is in a different section or something (also using storybook groups, etc.).

src/components/button-add-friend/ButtonAddFriend.ts
SharonStrats: I should probably wait until Timea merges that PR then so i can use the store she added.

src/components/button-add-friend/ButtonAddFriend.styles.css
SharonStrats: I've removed it.

src/components/button-add-friend/ButtonAddFriend.ts
SharonStrats: it depends on async work for refresh so i'm going to change it to state instead of a getter.

src/components/button-add-friend/ButtonAddFriend.ts
SharonStrats: sorry yes i understand now.

timea-solid:
I am not entirely sure what Noel was explaining.
What is the context here? What does this web component need to know?
Regarding the store - in all code we can get the store directly from solid-logic. In RDF forms we have a way to pass in a store for the sole purpose of using the RDF forms standalone, without solid-logic. So why would we need this option for this component? (This I never understood).

NoelDeMartin:
> I am not entirely sure what Noel was explaining.

With context I meant the `context` attribute for the component. Instead of an attribute, it should be a value provided with `@lit/context`. And yes, we should still be able to pass anything in case we want to override it (such as a store in the RDF Forms), but context (as in Lit context) should be the best practice.

That's just a suggestion, though, and how I think this should work... It's not impossible to pass the context directly as an attribute if that's the design choice we want to make. But I think it would be inconsistent to how we've been building all the other components.

SharonStrats:
@NoelDeMartin I added it as context, but then when I was trying to use it I ran into the decision on how the context should be provided. Are you able to share your vision about the providers we should have? 

We have the solid-ui-provider for instance that only has 'auth', should this also provide the store? Or should the component/pane that is using it have a provider for the store?

If the idea is that we should have the store on the solid-ui-provider, what is our intention with the DataBrowserContext from solid-logic/pane-registry? Should this become smaller or obsolete as we convert the panes to web components? 

I've written something up about the structure and redesign here https://sharon.pivot-test.solidproject.org:3000/public/MashlibDesign.md. I would love to hear your thoughts.

Note: when i created that document some research suggested that the solid logic store was stale, but when i dived deeper I don't think this is true. I think we can use the store in solid logic for the resource service to get content for the source pane. (I'll update this later) If you have an opinion about this I'd love to hear it.

NoelDeMartin:
> We have the solid-ui-provider for instance that only has 'auth', should this also provide the store? Or should the component/pane that is using it have a provider for the store?

Yes, I think it'd make sense to provide it in solid-ui-provider as well, that way consumers of solid-ui only need one wrapper to their application and it provides all the necessary contexts. Though this is the part that I was saying can depend on how we solve this for RDF forms. As far as I understand, RDF Form also need access to a global store, and ideally it should use the same that we use here.

> If the idea is that we should have the store on the solid-ui-provider, what is our intention with the DataBrowserContext from solid-logic/pane-registry? Should this become smaller or obsolete as we convert the panes to web components?

I'm not familiar with DataBrowserContext, but as far as I understand it, yes it should probably become obsolete in favour of Web Components. Everything we have in that context can probably be moved to an object provided by `@lit/context`.

PR #829, opened 2026-07-03
delete footer

### LINKED ISSUE #682: Implemnet re-redesign of logged in user menu items
We have a left side menu, when user logged in. We migth need to move this under the header top right.
Also the nav bar menu might move place.

timea-solid:
I am changing the layout according to new design. We do not have a footer anymore.


timea-solid:
make sure to also merge solid-panes and mashlib otherwise, CIs will break there. See parent ticket where all are linked.

README.md
```diff
@@ -216,20 +216,6 @@ I want this all to be presented flexible in the component.
 
 * Raptor mini: when we are on layout mobile we do not want to display the help menu at all.
 
-* Raptor mini: Create for me a footer Lit Component in tsy style of the components I have and under v2. Take the code from this index.ts to start with.
```
NoelDeMartin, line 230: Is it ok to remove these comments? They are still part of the "prompt history", even if the code they generated is now removed.

I don't like these here either, to be honest, but if we want to keep prompt history it should be exhaustive. Maybe we can just move the to a separate .md file outside of the README or something?

src/utils/headerHelpers.ts
```diff
@@ -0,0 +1,126 @@
+/*
+    Copied from mashlib/src/global/metadata.ts
```
NoelDeMartin: I wouldn't add this comment here. If you want to keep attribution, the commit message is a much better place for this.

What if we keep refactoring this file in the future? It won't be "copied from mashlib" anymore. Or at what point does it stop being so? After changing one line? Two? The entire file? etc.

src/utils/headerHelpers.ts
NoelDeMartin, line 1: Actually, I'm not sure we even want this file here. What's the purpose of moving it here? It's not even exported from any entrypoint, and isn't used internally either, so it's effectively dead code.

Also, if we wanted some "header helpers" I'd also suppose they should go in the package where the header is defined, `solid-panes`.

NoelDeMartin (changes requested):
The code changes are fine (just removing code, which is nice xD), but there are a couple of things I would change before merging this.

README.md
timea-solid: we should delete the prompts everywhere in the end. this approach was not ok. for now I deleted the prompts that I used to create the files I deleted.

src/utils/headerHelpers.ts
timea-solid: I did not mean to touch this file at all, this is an old code file used in the old header and footer believe. we will delete this when we depreciate the old header and footer. not now.

PR #833, opened 2026-07-09
#832 Implement initializing auth state

### LINKED ISSUE #832: Improve login UX
Currently, when the browser redirects back to the app after logging in, there is a brief period of time where the account status still appears as logged out. There are a couple of things we could improve here, such as adding a loading state to indicate that the app is still booting app, and try to reduce this time as much as possible.

NoelDeMartin:
This will show a spinner when authentication is still being initialized in the Account component, which should be helpful to distinguish whether you're looking at a logged-out page, or the authentication is still initializing.

It also opens the door to implementing more involved loading UIs, with the initialization state exposed in the Auth context and a new slot in the Guard component.

### DIFF
--- a/src/components/account/Account.stories.ts
+++ b/src/components/account/Account.stories.ts
@@ -15,12 +15,8 @@ const meta = {
 
 const render = defineAuthStoryRender(() => html`<solid-ui-account></solid-ui-account>`)
 
-export default meta
-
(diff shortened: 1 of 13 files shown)


timea-solid (approved):
I think this was a great improvement!

PR #837, opened 2026-07-17
deploy gh-pages from staging

bourgeoa:


NoelDeMartin (approved):
Only with this modification, the url for the storybook would be https://solidos.github.io/solid-ui/examples/storybook/ right? That is fine, though maybe it's a good time to rethink the url to something like `https://solidos.github.io/solid-ui/storybook/` directly (or maybe even `https://solidos.github.io/solid-ui/`). At this point, I'm not sure we even want to keep the html examples because I'm not sure they are maintaned... Even if they are, we can have both the storybook in the root and the examples in `/examples`.

Just a suggestion, though. We can move forward with this for now.

bourgeoa:
This could be part of the documentation step review..

PR #850, opened 2026-08-02
syncItems filter now excludes only slot="trigger" instead of all chil…

### LINKED ISSUE #849: solid-ui+menu - syncItems filters children without slot, then assigns them slot and get filtered out
There is a bug in solid-ui
> It's a bug in solid-ui-menu: syncItems filters children without slot, then assigns them slot — so the second time it runs, all children have a slot and get filtered out. Let me work around it by using <wa-dropdown> directly:

bourgeoa:
…dren with any slot. First run assigns slots, second run keeps them — no more empty menu

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2161,20 +2161,22 @@
 			}
 		},
 		"node_modules/@emnapi/core": {
-			"version": "2.0.0-alpha.3",
-			"resolved": "https://registry.npmjs.org/@emnapi/core/-/core-2.0.0-alpha.3.tgz",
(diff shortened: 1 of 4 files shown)


SharonStrats (commented):
I think this makes sense even if we don't need it in this specific thing (as I think we should solve this case by just rerendering the pane itself not the header) but we may run into this again in the future.  I just want to test it out with the panes that are using it though to make sure it doesn't effect them. I will do it soon. I just want to finish this editing task first before changing my environment to test.

SharonStrats (approved):
Looks good! I think we need this no matter what.

PR #853, opened 2026-08-06
Add theme options for code editor

SharonStrats:
This PR is against another PR where I moved the code editor to solid-ui (not staging yet). @bourgeoa you mentioned not loving the dark theme so I thought I'd just have a quick go at seeing what our options were for allowing users to change it inside the editor so we don't have to store their preferences (since this would mean deciding on a predicate etc and may take longer to implement)

Here is how it looks, but we could easily change the look this is just a rough draft. I created this for discussion. 
Note: the light theme can be changed as well there are many different extensions we could pick from, if we did do this and chose this one i'd put some sort of border around the code.

Screenshot 2026-08-06 at 11 14 54 AM

Screenshot 2026-08-06 at 11 22 19 AM

### DIFF
--- a/src/lib/code-editor/CodeEditor.styles.css
+++ b/src/lib/code-editor/CodeEditor.styles.css
@@ -1,0 +1,52 @@
+.code-editor-shell {
+	display: flex;
+	flex-direction: column;
+	gap: 0.5rem;
+	width: 100%;
+	min-width: 0;
+}
+
+.code-editor-toolbar {
+	display: flex;
+	align-items: center;
+	justify-content: flex-end;
+	gap: 0.5rem;
+	flex-wrap: wrap;
+	width: 100%;
+	min-width: 0;
+}
+
+.code-editor-toolbar-label {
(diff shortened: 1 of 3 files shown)


bourgeoa:
Thanks for looking at it..
I'm not very found of your implementation because it adds a line. We already have 2 which I think is one to much. We must limit us to one line per Solid Resource.

As for the source-pane theme I think it should be global with the other themes parameters.

SharonStrats:
> Thanks for looking at it.. I'm not very found of your implementation because it adds a line. We already have 2 which I think is one to much. We must limit us to one line per Solid Resource.
> 
> As for the source-pane theme I think it should be global with the other themes parameters.

ok I will close this ticket and create an issue for adding a preference for this in User Preferences. I agree this is the best way especially long term. 

Just a note about the double lines. This is because of the container above is showing the same as the non container resource. This will go away when I finish off the file explorer design. the non container resources just show by themselves and just the resource header. the container will show off to the left side closed (instead of above as it is showing now).

PR #868, opened 2026-09-12
add refresh metadata

SharonStrats:
This was missing from the restore session flow, it only happens on the login flow.

### DIFF
--- a/src/lib/auth/sessionRestore.ts
+++ b/src/lib/auth/sessionRestore.ts
@@ -1,0 +1,15 @@
+import { authSession, solidLogicSingleton } from 'solid-logic'
+
+let restoreListenerRegistered = false
+
+export function registerAuthorizationMetadataInvalidationOnSessionRestore () {
+  if (restoreListenerRegistered) {
+    return
+  }
+
+  authSession.events.on('sessionRestore', () => {
+    solidLogicSingleton.store.updater.flagAuthorizationMetadata() as any
+  })
+
+  restoreListenerRegistered = true
+}
--- a/src/login/login.ts
+++ b/src/login/login.ts
@@ -47,6 +47,7 @@ import { style } from '../lib/style'
 import { alert } from '../lib/log'
 import ns from '../lib/ns'
 import { Signup } from '../signup/signup.js'
+import { registerAuthorizationMetadataInvalidationOnSessionRestore } from '../lib/auth/sessionRestore'
 import * as utils from '../utils'
 import * as widgets from '../widgets'
 
@@ -741,6 +742,8 @@ authSession.events.on('logout', async () => {
   }
   window.location.reload()
 })
+
+registerAuthorizationMetadataInvalidationOnSessionRestore()
 
 /**
  * Workspace selection etc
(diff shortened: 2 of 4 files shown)


src/login/login.ts
```diff
@@ -742,6 +743,8 @@ authSession.events.on('logout', async () => {
   window.location.reload()
 })
 
+registerAuthorizationMetadataInvalidationOnSessionRestore()
```
NoelDeMartin, line 746: I'm not sure what this is supposed to do (haven't looked into solid-logic), but I'm pretty sure this shouldn't go here. In particular, this is creating a "side-effect", meaning that this function will be called just by importing this file. Instead, the function should be called explicitly when necessary. For example, as it's already being called in the `initialize` method of SolidAuth. Shouldn't that be sufficient?

src/lib/auth/sessionRestore.ts
```diff
@@ -0,0 +1,15 @@
+import { authSession, solidLogicSingleton } from 'solid-logic'
+
+let restoreListenerRegistered = false
+
+export function registerAuthorizationMetadataInvalidationOnSessionRestore () {
```
NoelDeMartin, line 5: It seems like everything this helper is doing is registering a listener that updates something in the `solid-logic` singleton. However, given that both the event and the implementation are only interacting with `solid-logic` code, I wonder why this function is defined in this project? Furthermore, I'm not sure why it's being called from this project at all. `solid-logic` already has all the implementation for session management, right? Why don't we handle this there?

NoelDeMartin (changes requested):

src/lib/auth/sessionRestore.ts
SharonStrats: Ok, I think I should be able to move it to solid-logic. Maybe in the auth... I'll look into that further.

src/login/login.ts
SharonStrats: I could just add the listener here, that is what I had before the copilot Review. The event listener though is crucial otherwise we can not refresh a resource. You see it's called here https://github.com/SolidOS/solid-ui/blob/5881f39ff1bbb26b5e4c006ce31c0fc2354d8401/src/login/login.ts#L722 on logout. This is very important.

Do you think I should just put it back to authSession.events.on('sessionRestore', () => {
    solidLogicSingleton.store.updater.flagAuthorizationMetadata() as any
  }) and not touch solidAuth?

NoelDeMartin: I wasn't aware of that part of the code, but looking at that, I think that isn't correct either because it's also a side-effect 😅. Though I guess it's better to add it there than create a new side-effect. But I think we should also remove that whole `authSession.events.on('logout', async () => {` block from this file.

SharonStrats: do you have any suggestions on where it should go or what I should research?

I also thought login code was not being used, but it definitely is. I am not sure why. I can research why... but are you able to give me any guidance about what you think this should look like? Has not everything for the header code been moved to web components by chance?

SharonStrats: Okay after doing more research I understand what you are saying now.

SharonStrats:
This was not the correct approach.

PR #869, opened 2026-09-14
New layout playground

SharonStrats:


SharonStrats:
@NoelDeMartin so here I added to the solid-ui-provider. I also extended the menu to allow me to change the font size and gap because the 3 dot menu has different styling. And oops I'm tired 😅 I just realized this is where I have moved the file explorer web components so I can use them in both solid-panes and folder-pane.

Highlevel for providers we have
solid-ui-provider. (auth, and I thought we should add store here which I've done)
file-explorer-provider
storage-pane-provider.  (and source-pane-provider) sitting at this same level

src/components/file-explorer-header/FileExplorerProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+      changedProperties.has('deleteResource') ||
+      changedProperties.has('resourceRevision') ||
+      changedProperties.has('paneSupportsEditing') ||
+      changedProperties.has('isDirty')
+    ) {
+      this.refreshFileExplorerContextValue()
+    }
+  }
+
+  render () {
```
NoelDeMartin, line 255: "provider" components should always be headless, meaning that they only render a `<slot>` for their children. For example, that's how `<solid-ui-provider>` works (It does also render the dialogs root, but that's fine since those are absolutely positioned).

src/components/file-explorer-header/index.ts
```diff
@@ -0,0 +1,5 @@
+import FileExplorerHeader from './FileExplorerHeader'
+import FileExplorerProvider from './FileExplorerProvider'
+
+export { FileExplorerHeader, FileExplorerProvider }
```
NoelDeMartin, line 4: This folder is exporting 2 components, but if those 2 are supposed to be "public", they should each have their own folder.

However, looking at how all the file-explorer is structured, I wonder if this needs to be within solid-ui at all? Shouldn't it be a pane instead?

Either way, if this really ends up as a component in solid-ui, I think it should just be `<solid-ui-file-explorer>`, and all the other components and helpers should be internal to the package.

src/components/resource-actions-menu/ResourceActionsMenu.ts
```diff
@@ -0,0 +1,111 @@
...
+        return html`<icon-lucide-trash-2 slot="left-icon"></icon-lucide-trash-2>`
+      }
+
+      return nothing
+    }
+
+    return html`
+      <solid-ui-menu>
+        <solid-ui-button slot="trigger" variant="ghost" title="More options">
+          <icon-lucide-ellipsis-vertical slot="icon" class="ellipsisIcon"></icon-lucide-ellipsis-vertical>
```
NoelDeMartin, line 100: I have seen this mistake in a lot of places during my review, so I won't address all of them. But in general, it's not a good a11y practice to have buttons without text content. I realize this button has a "title", but many accessibility tools ignore that attribute completely. Furthermore, even using `aria-label` is not ideal because translation tools often ignore it. The best approach is to simply add a `<span class="sr-only">` inside of the button. It's ok if the text is the same as the title, those serve different purposes (one is for visually impaired users, the other one is for mouse users hovering the button).

src/components/file-explorer-header/FileExplorerProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  private async getPaneItems (subject: NamedNode, context: DataBrowserContext, relevantPanes: PaneDefinition[]) {
+    const dom = context.dom
+    const menuItems = await Promise.all(relevantPanes.map(async pane => {
+      const label = pane.label(subject, context) ?? ''
+
+      let icon: HTMLElement | undefined
+      if (pane.name === 'profile') {
+        icon = dom.createElement('icon-lucide-user')
```
NoelDeMartin, line 169: We should avoid using this imperative API. Instead, the icons can be rendered like this:

```js
html`<icon-lucide-user slot="left-icon"></icon-lucide-user>`
```

Other than that, I don't see `~icons/lucide/user` being imported in this file. Components should always be imported where they are declared (same for the `users` icon down below, etc.).

src/components/file-explorer-header/FileExplorerProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    const subject = store.sym(this.subjectUri)
+    return html`
+      <div class="file-explorer-provider">
+        ${this.showHeader
+          ? html`
+              <file-explorer-header
+                .paneIcon=${this.getPaneIcon(this.pane, subject, this.context)}
+                .menuItems=${this.menuItems}
```
NoelDeMartin, line 268: I think this should have a different name, maybe `extraMenuItems`? It seems like the file explorer already has some built-in menu items that cannot be controlled by this property. So this actually only serves to add some additional entries.

src/components/file-explorer-header/FileExplorerProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+                .paneSupportsEditing=${this.paneSupportsEditing}
```
NoelDeMartin, line 269: Maybe these 3 should be passed down to the components using Lit's context, rather than properties? I assume the `<file-explorer-header>` is not going to be used on its own.

src/components/file-explorer-header/FileExplorerProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+  accessor resourceRevision: number | undefined = undefined
+
+  @property({ attribute: false })
+  accessor soloPane: boolean | undefined = undefined
+
+  @property({ attribute: false })
+  accessor openPane: ((subject: NamedNode, paneName: string) => void) | undefined = undefined
+
+  @consume({ context: fileExplorerContext, subscribe: true })
+  accessor parentFileExplorerContext: FileExplorerContext = undefined as unknown as FileExplorerContext
```
NoelDeMartin, line 86: This is something we should avoid doing, because it silences one potential bug. This assumes that a `FileExplorerContext` is provided by a parent component, but if it's missing it'll just throw a "trying to use `.foobar` from undefined" or something, instead of saying "file explorer context is missing".

There are 2 ways to solve this:

- Provide a default noop context (this is how auth and context stores are used in other components, for example).
- Declare this as `FileExplorerContext | undefined`, and handle all the situations where a missing context would be an issue (throwing an error or doing something else).

I actually prefer the first, and looking at this PR I thought of a way to have better DX. I'm going to add a comment on the `NoopStore` about it.

src/lib/store/NoopStore.ts
```diff
@@ -0,0 +1,19 @@
+import { LiveStore } from 'rdflib'
+
+const throwNoopStoreError = () => {
+  throw new Error('Cannot use RDF forms without a store')
+}
+
+const NoopStore = new Proxy({} as LiveStore, {
```
NoelDeMartin, line 7: I really like this idea, but I don't think it should be something specific to the Store context, we could use this for all the `DEFAULT_*` context we have been using (Auth, Store, FileExplorerContext, etc.)

We could create a generic helper like this:

```ts
export function noopContext<T>(name: string): T {
    function throwNoopError() {
        throw new Error(`Context for '${name}' is missing.`);
    }

    return new Proxy({}, {
        get: throwNoopError,
        // ...
    }) as unknown as T;
}
```

And then just use it like this everywhere that we require a context:

```ts
  @consume({ context: authContext, subscribe: true })
  private accessor auth = noopContext<AuthContext>('auth');
```

src/components/resource-actions-menu/ResourceActionsMenu.ts
```diff
@@ -0,0 +1,111 @@
...
+
+@customElement('solid-ui-resource-actions-menu')
+export default class ResourceActionsMenu extends WebComponent {
+  static styles = styles
+
+  @property({ type: Boolean })
+  accessor isContainerResource = false
+
+  @property({ type: Boolean })
+  accessor isMobile = false
```
NoelDeMartin, line 25: I noticed that this and other components are re-implementing this sort of functionality. Instead of doing this, we should rely on the `breakpoints.css` variables (see how those are used in the `Header` component in solid-panes, for example).

If CSS breakpoints are not sufficient, we could create a UI context, similar to Auth, Store, etc. that is provided by `<solid-ui-provider>` and implements this reusable functionality. That way, we would have all the breakpoints and responsive functionality encapsulated.

src/components/file-explorer-header/FileExplorerHeaderControls.ts
```diff
@@ -0,0 +1,206 @@
...
+export default class FileExplorerHeaderControls extends WebComponent {
+  static styles = styles
+
+  private mobileMediaQuery: MediaQueryList | undefined
+  private readonly mobileQuery = '(max-width: 600px)'
+  private readonly handleMobileMediaChange = (event: MediaQueryListEvent) => {
+    this.isMobile = event.matches
+  }
+
+  private handleMobileAccessClick = () => {
```
NoelDeMartin, line 27: This is more of a pet peeve of mine, nothing technically wrong with this, but I think it's always better to define private methods rather than variable methods. (also move this down in the file, the first things we should see are the properties, etc.).

```suggestion
  private handleMobileAccessClick() {
```

src/components/resource-actions-menu/ResourceActionsMenu.ts
```diff
@@ -0,0 +1,111 @@
...
+      handleEditingClick: this.handleEditingClick,
+      handleDeleteClick: this.handleDeleteClick,
+      discoverPublicly: this.discoverPublicly,
+      discoverPrivately: this.discoverPrivately,
+      handleDiscoverPublicClick: this.handleDiscoverPublicClick,
+      handleDiscoverPrivateClick: this.handleDiscoverPrivateClick,
+      menuItems: this.menuItems,
+    })
+
+    const renderIcon = (item: ResourceActionMenuItem) => {
```
NoelDeMartin, line 77: This could be defined as a private method instead of an inline function, to avoid re-declaring it on every render.

src/components/menu-item/MenuItem.styles.css
```diff
@@ -1,17 +1,20 @@
...
         justify-content: flex-start;
         align-items: center;
-        gap: 5px;
+        gap: var(--solid-ui-menu-item-gap, 5px);
         padding: 10px;
         border-radius: 5px;
         color: var(--solid-ui-color-gray-600);
         font-weight: 500;
-        font-size: var(--solid-ui-font-size-md);
+        font-size: var(--solid-ui-menu-item-font-size, var(--solid-ui-font-size-md));
```
NoelDeMartin, line 17: I noticed we're adding some modifications here, and overriding a lot of the styles for the menu in other parts of the code. This should be avoided if possible, the point of the Design System is that we reuse component and styles. If we want something to be customizable, it should be with component properties. For example, if we want menus with bigger or smaller font sizes, we could have a `size` property on the top `<solid-ui-menu>` component.

src/components/resource-actions-menu/ResourceActionsMenu.styles.css
```diff
@@ -0,0 +1,36 @@
...
+    --solid-ui-menu-min-width: fit-content;
+  }
+
+  .ellipsisIcon {
+    width: 16px;
+    height: 16px;
+  }
+}
+
+solid-ui-menu::part(menu) {
```
NoelDeMartin, line 17: As mentioned in the other comment, we should override Design System styles so aggressively. It's ok to override some things in some edge cases or something, but if you're redefining the styles of the component completely, you probably need to fix the underlying component instead (or maybe create a new component altogether).

src/components/resource-actions-menu/helpers.ts
```diff
@@ -0,0 +1,75 @@
...
+  handleEditingClick?: (() => void) | undefined
+  handleDeleteClick?: (() => void) | undefined
+  discoverPublicly?: boolean
+  discoverPrivately?: boolean
+  handleDiscoverPublicClick?: (() => void) | undefined
+  handleDiscoverPrivateClick?: (() => void) | undefined
+  menuItems?: ResourceActionMenuItem[]
+}
+
+export function getVisibleResourceActions (options: ResourceActionsMenuOptions): ResourceActionMenuItem[] {
```
NoelDeMartin, line 24: Why is this exported to the public and in a separate file? Couldn't it be just a private method in the component?

src/lib/discovery/discovery.ts
```diff
@@ -0,0 +1,45 @@
...
+export const DEFAULT_DISCOVER_CLASS = ns.owl('Thing')
+
+export type DiscoveryVisibility = 'public' | 'private'
+
+export type DiscoveryState = {
+  public: boolean
+  private: boolean
+}
+
+export async function loadDiscoveryState (subject: NamedNode): Promise<DiscoveryState> {
```
NoelDeMartin, line 14: I'm not sure why this and the `resource-actions-menu.ts` helpers are exported from the library, and even declared outside of the component folders. Shouldn't these be internal helpers for the file explorer components?

src/lib/file-explorer/relevant-panes.ts
```diff
@@ -0,0 +1,42 @@
...
+import { filterAvailablePanes } from '../../login/login'
+import type { NamedNode } from 'rdflib'
+
+export async function getRelevantPanes (
+  subject: NamedNode,
+  context: DataBrowserContext,
+  panes: Array<PaneDefinition> = context.session.paneRegistry.list
+): Promise<Array<PaneDefinition>> {
+  const relevantPaneCandidates = panes.filter(
+    (pane) => pane.label(subject, context) && !pane.global && pane.name !== 'sharing'
```
NoelDeMartin, line 11: I'm not sure this is the best way to get the relevant pane, what if a pane has a fallback label for unknown subjects such as "Unknown"? I'm not super familiar with how components are resolved, or how do we decide which pane should be used to render a certain subject. But I think it should be more robust that just looking for an empty label.

NoelDeMartin (commented):
I have added a bunch of comments about specific things to improve, but in general I think it's looking good and using the Design System components correctly :).

The only thing I wonder is whether the file explorer should really be a component from solid-ui, or a dedicated pane. Either way, I think everything should be as encapsulated as possible. In general, I think it's better to expose as little public API as possible and just expose helpers or reusable functionality as necessary. Otherwise, you risk breaking dependencies if you later want to refactor something that should have been internal.

src/components/file-explorer-header/FileExplorerProvider.ts
SharonStrats: Ahhhh I did not know that.

src/components/file-explorer-header/index.ts
SharonStrats: Oh ok, I'll do that. It has to be in solid-ui because both solid-panes and folder-pane use them.

src/components/menu-item/MenuItem.styles.css
SharonStrats: Ahh Ok I put it in the wrong place.. makes sense.

src/components/resource-actions-menu/ResourceActionsMenu.ts
SharonStrats: thanks I'll have a look. I didn't see breakpoints.css. This I just through in to get it working but I'll be careful when I do it for real.

src/lib/discovery/discovery.ts
SharonStrats: I'll take a look at what you are saying. I had to have this in 2 places folder-pane and solid-panes if I remember right htat is why, but I'll see if I can do that better.

src/lib/file-explorer/relevant-panes.ts
SharonStrats: I will have a look. I took this code how it was.

